PR #14, opened 2017-11-24
Feature/setup domain model

kacacukurov:


core/src/main/java/eu/execom/testgenerator/core/repository/AnswerRepository.java
```diff
@@ -0,0 +1,10 @@
+package eu.execom.testgenerator.core.repository;
+
+import eu.execom.testgenerator.core.model.Answer;
+import org.springframework.data.jpa.repository.JpaRepository;
+
+/**
+ * Created by Katarina Cukurov on 24/11/2017.
```
stanislavgrujic: Please do not use this kind of JavaDoc, since it becomes obsolete very soon, and does not bring too much information. Apply on all files where it was added.

core/src/main/resources/db/changelog/db.changelog-createAnswerTable.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet author="kaca" id="2017-11-24 13:40">
```
stanislavgrujic: Please use timestamp instead of a date for an id.

core/src/main/java/eu/execom/testgenerator/core/model/Question.java
```diff
@@ -0,0 +1,50 @@
...
+  @NotNull
+  private boolean isActive;
+
+  @Column(nullable = false)
+  @NotNull
+  private int points;
+
+  @Column(nullable = false)
+  @NotNull
+  private int type;
```
stanislavgrujic: Please introduce an enumeration to represent a type, instead of using an int (e.g.
```
public enum QuestionType {

  FREE_FORM,
  MULTIPLE_CHOICE,
  // etc.
}
```

When saving to the database, use the enum value, not ordinal number.

core/src/main/java/eu/execom/testgenerator/core/model/Recipe.java
```diff
@@ -0,0 +1,32 @@
...
+
+  @ManyToOne
+  private Formula formula;
+
+  @Column(nullable = false)
+  @NotNull
+  private int numberOfQuestions;
+
+  @ManyToOne
+  private Category category;
```
stanislavgrujic: Please switch places with this field and `numberOfQuestions`, it will be easier to understand the connection between the entities if they are nicely grouped in the code.

core/src/main/java/eu/execom/testgenerator/core/model/User.java
```diff
@@ -0,0 +1,41 @@
...
+  private String name;
+
+  @Column(nullable = false)
+  @NotNull
+  @Email
+  private String email;
+
+  @Column(nullable = false)
+  @NotNull
+  private int role;
```
stanislavgrujic: Please introduce an enum to represent a user role.

core/src/main/resources/db/changelog/db.changelog-initializeDatabaseSchema.xml
```diff
@@ -0,0 +1,233 @@
...
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <property name="now" value="sysdate" dbms="oracle"/>
+    <property name="now" value="now()" dbms="mysql"/>
+
+    <changeSet author="kaca" id="${now}">
+        <createTable tableName="user">
```
stanislavgrujic: Would be nice to also have a comment in each `changeSet` explaining what was done in it. In this case it's pretty clear, but we want to have it established as a sample for further changes.
e.g.
```
<comment>Initial schema</comment>
<createTable tableName="user"> 
```

core/src/main/resources/db/changelog/db.changelog-initializeDatabaseSchema.xml
```diff
@@ -0,0 +1,233 @@
...
+            </column>
+        </createTable>
+
+        <addForeignKeyConstraint baseColumnNames="user_id"
+                                 baseTableName="test"
+                                 constraintName="fk_user_test"
+                                 onDelete="CASCADE"
+                                 onUpdate="RESTRICT"
+                                 referencedColumnNames="id"
+                                 referencedTableName="user"/>
```
stanislavgrujic: Could you please reorder the foreignKeyConstraint definition attributes as the following:
```
<addForeignKeyConstraint constraintName="fk_user_test"
                         baseTableName="test"
                         baseColumnNames="user_id"
                         referencedTableName="user"
                         referencedColumnNames="id"
                         onDelete="CASCADE"
                         onUpdate="RESTRICT"/>
```

Ordering by table -> id and grouping the column from the table and the referenced column makes it easier to figure out the connection.
Please apply on all foreign key constraint definitions.

core/src/main/java/eu/execom/testgenerator/core/model/Test.java
```diff
@@ -0,0 +1,37 @@
...
+  @Id
+  @NotNull
+  private Long id;
+
+  @ManyToOne
+  @NotNull
+  private User user;
+
+  @ManyToOne
+  private User review;
```
stanislavgrujic: Please rename to `reviewer` (also in the database).

core/src/main/java/eu/execom/testgenerator/core/model/Test.java
```diff
@@ -0,0 +1,37 @@
...
+@Data
+public class Test {
+
+  @Id
+  @NotNull
+  private Long id;
+
+  @ManyToOne
+  @NotNull
+  private User user;
```
stanislavgrujic: Please rename to `applicant` (also in the database).

Issue #3, opened 2017-11-02
Create User with Google OAuth


stanislavgrujic:

millevlada:
issue closed

PR #15, opened 2017-11-27
Added creating user with Google OAuth. (#3)

kacacukurov:


core/src/main/java/eu/execom/testgenerator/core/service/UserServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+public class UserServiceImpl implements UserService{
+
+  @Autowired
+  private UserRepository userRepository;
+
+  @Override
+  @Transactional(readOnly = false)
```
stanislavgrujic: `readOnly` has default value `false`, so no need to explicitly put it here. Please remove it.

core/src/main/java/eu/execom/testgenerator/core/service/UserServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+  public User save(User user) {
+    return this.userRepository.save(user);
+  }
+
+  @Override
+  @Transactional(readOnly = false)
```
stanislavgrujic: Please remove `readOnly`, it is `false`, by default.

web/src/main/java/eu/execom/testgenerator/web/WebComponents.java
```diff
@@ -0,0 +1,5 @@
+package eu.execom.testgenerator.web;
+
+//marker interface
```
stanislavgrujic: Please update the comment appropriately.

web/src/main/java/eu/execom/testgenerator/web/controller/UserController.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @RequestMapping(
+      value = "/register",
+      method = RequestMethod.GET,
+      produces = MediaType.APPLICATION_JSON_VALUE)
+  public Principal register(Principal principal) {
+    OAuth2Authentication oAuth2Authentication = (OAuth2Authentication) principal;
+    Authentication authentication = oAuth2Authentication.getUserAuthentication();
+    Map<String, String> details = new LinkedHashMap<>();
+    details = (Map<String, String>) authentication.getDetails();
```
stanislavgrujic: No need to create one variable and then do assignment in another line, you can merge the two lines above:
```
Map<String, String> details = (Map<String, String>) authentication.getDetails();
```

web/src/main/java/eu/execom/testgenerator/web/controller/UserController.java
```diff
@@ -0,0 +1,48 @@
...
+
+    String email = details.get("email");
+    if(email.contains("@execom.eu")){
```
stanislavgrujic: Please check you formatter, it seems as if it is not working properly.

stanislavgrujic: Also, please extract the string as constant with a name EXECOM_DOMAIN and update the condition to
```
if (email.endsWith(EXECOM_DOMAIN)) {
```

web/src/main/java/eu/execom/testgenerator/web/controller/UserController.java
```diff
@@ -0,0 +1,48 @@
...
+      if(userService.findByEmail(email) == null){
+        User user = new User();
+        user.setEmail(email);
+        user.setName(details.get("name"));
+        user.setRole(UserRole.ADMINISTRATOR);
+        userService.save(user);
+      }
+    }
+    System.out.println(details.toString());
```
stanislavgrujic: You should use a proper logger, instead of printing to console. I will create a story for that.

pom.xml
```diff
@@ -23,6 +23,8 @@
     <properties>
     <jdk.version>1.8</jdk.version>
     <lombok.version>1.16.18</lombok.version>
+    <slf4j-api.vesrion>1.7.25</slf4j-api.vesrion>
+    <slf4j-log4j12>1.7.25</slf4j-log4j12>
```
stanislavgrujic: This should be one property for both dependencies.

pom.xml
```diff
@@ -23,6 +23,8 @@
     <properties>
     <jdk.version>1.8</jdk.version>
     <lombok.version>1.16.18</lombok.version>
+    <slf4j-api.vesrion>1.7.25</slf4j-api.vesrion>
```
stanislavgrujic: Please fix spelling and replace these two properties with: `slf4j-version`.

web/src/main/java/eu/execom/testgenerator/web/controller/UserController.java
```diff
@@ -22,6 +25,8 @@
 
   private static final String EXECOM_DOMAIN = "execom.ue";
 
+ private final Logger logger = LoggerFactory.getLogger(UserController.class);
```
stanislavgrujic: Please use Lombok annotation instead `@Slf4j`

web/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,12 @@
...
+# Define the file appender
+log4j.appender.file=org.apache.log4j.RollingFileAppender
+log4j.appender.file.File=log/log4j.log
+log4j.appender.file.layout = org.apache.log4j.PatternLayout
+log4j.appender.file.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
+
+# Define the console appender
+log4j.appender.stdout=org.apache.log4j.ConsoleAppender
+log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
+log4j.appender.stdout.layout.ConversionPattern=%d{ABSOLUTE} %5p %c:%L - %m%n
```
stanislavgrujic: Could you please use the same Conversion pattern for both appenders?

core/src/main/java/eu/execom/testgenerator/core/model/Recipe.java
```diff
@@ -5,10 +5,13 @@
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.ManyToOne;
 import javax.persistence.Table;
 import javax.validation.constraints.NotNull;
+import java.text.Normalizer;
```
stanislavgrujic: Why do you need this?

core/src/main/java/eu/execom/testgenerator/core/service/UserServiceImpl.java
```diff
@@ -0,0 +1,26 @@
+package eu.execom.testgenerator.core.service;
+
+import eu.execom.testgenerator.core.model.User;
+import eu.execom.testgenerator.core.repository.UserRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+public class UserServiceImpl implements UserService{
```
stanislavgrujic: Please make this class package protected - we do not want to have it exposed directly in the web module, only through the interface.

core/src/main/resources/db/changelog/db.changelog-master.xml
```diff
@@ -5,6 +5,6 @@
          http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
 
-    <include file="classpath:/db/changelog/db.changelog-initializeDatabaseSchema.xml"/>
+<!--    <include file="classpath:/db/changelog/db.changelog-initializeDatabaseSchema.xml"/> -->
```
stanislavgrujic: Could you uncomment this line?

PR #18, opened 2017-12-04
Feature/create service questions crud

kacacukurov:


core/src/main/java/eu/execom/testgenerator/core/service/QuestionService.java
```diff
@@ -0,0 +1,14 @@
...
+
+public interface QuestionService {
+
+  Question save(Question question);
+
+  Question findOne(Long id);
+
+  Question update(Question question);
+
+  void remove(Long id);
```
stanislavgrujic: Please also add method to find all questions which belong to a Category.

Issue #11, opened 2017-11-08
Create service to manage User CRUD


stanislavgrujic:

millevlada:
issue closed

PR #19, opened 2017-12-06
Added new methods to userService. (#11)

kacacukurov:


core/src/main/java/eu/execom/testgenerator/core/service/UserServiceImpl.java
```diff
@@ -15,12 +15,30 @@
...
 
   @Override
   @Transactional
   public User findByEmail(String email){
-    return this.userRepository.findByEmail(email);
+    return userRepository.findByEmail(email);
+  }
+
+  @Override
+  @Transactional
```
stanislavgrujic: Please add `readOnly = true`, on all methods where we don't update the database.

PR #32, opened 2017-12-14
Created TestController, added/updated necessary methods in TestService

luinnarn:


web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,63 @@
...
+
+@RestController
+public class TestController {
+
+    @Autowired
+    private TestService testService;
+
+    @RequestMapping(value = "/createtest/{formulaId}/{userId}",
+                    method = RequestMethod.POST)
+    public ResponseEntity<TestDto> createTest(@PathVariable("formulaId") Long formulaId,
```
stanislavgrujic: Please make the path `/test/create` and pass `formulaId` and `userId` as part of the request body.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,63 @@
...
+    public ResponseEntity<TestDto> getTestById(@PathVariable("testId") Long testId) {
+        Test test = testService.findOne(testId);
+        TestDto dto = new TestDto(test);
+        return new ResponseEntity<TestDto>(dto, HttpStatus.OK);
+    }
+
+    @RequestMapping(value = "/tests",
+                    method = RequestMethod.GET)
+    public ResponseEntity<List<TestDto>> getTests() {
+        List<Test> testList = testService.findAll();
```
stanislavgrujic: Please rename to `tests` - it is not a good practice to add type info to the variable name.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,63 @@
...
+        List<TestDto> testDtoList = new ArrayList<TestDto>();
+        for (Test test : testList) {
+            testDtoList.add(new TestDto(test));
+        }
```
stanislavgrujic: We are using Java 8 in the project. Feel free to get to know it a bit, if you are unfamiliar with its features. :)

The upper 4 lines can be written as:
```
List<TestDto> testDtos = tests.stream().map(TestDto::new).collect(Collectors.toList());
```

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,63 @@
...
+        return new ResponseEntity<List<TestDto>>(testDtoList, HttpStatus.OK);
+    }
+
+    @RequestMapping(value = "/users/{userId}/tests",
```
stanislavgrujic: Please change mapping to `/tests/{userId}`

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,63 @@
...
+                    method = RequestMethod.GET)
+    public ResponseEntity<List<TestDto>> getTestsByUserId(@PathVariable("userId") Long userId) {
+        List<Test> testList = testService.findTestsByUserId(userId);
+        List<TestDto> testDtoList = new ArrayList<TestDto>();
+        for (Test test : testList) {
+            testDtoList.add(new TestDto(test));
+        }
```
stanislavgrujic: See my comment above about replacing the upper 4 lines.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,63 @@
...
+        return new ResponseEntity<List<TestDto>>(testDtoList, HttpStatus.OK);
```
stanislavgrujic: Please use diamond operator instead of explicitly stating List type, e.g.,
```
return new ResponseEntity<List<>>(testDtoList, HttpStatus.OK);
```

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -75,6 +84,10 @@ public Test findOne(Long id) {
     return testRepository.findAll();
   }
 
+  @Override
+  @Transactional(readOnly = true)
+  public List<Test> findTestsByUserId(Long userId) { return userRepository.getOne(userId).getTest(); }
```
stanislavgrujic: Please do not inline method body with method header.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,63 @@
...
+        TestDto dto = new TestDto(test);
+        return new ResponseEntity<TestDto>(dto, HttpStatus.OK);
+    }
+
+    @RequestMapping(value = "/tests/{testId}",
+                    method = RequestMethod.GET)
+    public ResponseEntity<TestDto> getTestById(@PathVariable("testId") Long testId) {
+        Test test = testService.findOne(testId);
+        TestDto dto = new TestDto(test);
+        return new ResponseEntity<TestDto>(dto, HttpStatus.OK);
```
stanislavgrujic: Please use diamond operator instead.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,63 @@
...
+                    method = RequestMethod.GET)
+    public ResponseEntity<List<TestDto>> getTestsByUserId(@PathVariable("userId") Long userId) {
+        List<Test> testList = testService.findTestsByUserId(userId);
+        List<TestDto> testDtoList = new ArrayList<TestDto>();
+        for (Test test : testList) {
+            testDtoList.add(new TestDto(test));
+        }
+        return new ResponseEntity<List<TestDto>>(testDtoList, HttpStatus.OK);
+    }
+}
```
stanislavgrujic: Please format the file.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,63 @@
...
+import eu.execom.testgenerator.core.model.Test;
+import eu.execom.testgenerator.core.model.User;
+import eu.execom.testgenerator.core.service.FormulaService;
+import eu.execom.testgenerator.core.service.UserService;
+import eu.execom.testgenerator.web.dto.TestDto;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.method.P;
+import org.springframework.web.bind.annotation.*;
```
stanislavgrujic: Please do not use group imports. Setup your IDE to group imports only if there are more than 50 from the same package.

web/src/main/java/eu/execom/testgenerator/web/dto/TestDto.java
```diff
@@ -0,0 +1,17 @@
...
+
+@NoArgsConstructor
+@Data
+public class TestDto {
+    private String applicantName;
+    private String testQuestions;
+
+    public TestDto(Test test) {
+        this.applicantName = test.getApplicant().getName();
+        this.testQuestions = test.getTestQuestions().toString();
```
stanislavgrujic: Please make this similar to the `Test` class, just without the information about the database mapping and use id's instead of whole classes references.

web/src/main/java/eu/execom/testgenerator/web/dto/TestResponse.java
```diff
@@ -16,9 +20,14 @@
     private Long reviewerId;
     private Long formulaId;
 
-
     public TestResponse(Test test) {
         this.applicantName = test.getApplicant().getName();
-        this.testQuestions = test.getTestQuestions().toString();
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            this.testQuestions = mapper.writeValueAsString(
```
stanislavgrujic: Please do not serialize manually, Jackson does that for you automatically when you send back the response.

web/src/main/java/eu/execom/testgenerator/web/dto/TestQuestionDto.java
```diff
@@ -0,0 +1,17 @@
+package eu.execom.testgenerator.web.dto;
+
+import eu.execom.testgenerator.core.model.TestQuestion;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
+@Data
+public class TestQuestionDto {
```
stanislavgrujic: TestQuestion is a structure introduced to represent a join table between Test and a Question. When you send the information to the UI (or wherever) through controller, you just need the information from Question, so, you should write your TestResponse class similar to this:
```
public class TestResponse {
  // everything else
  private List<QuestionDto> questions;
}

public class QuestionDto {
  
  private Long id;
  private QuestionType type;
  private String text;
  // investigate what is the best way to send an image to the UI
  private byte[] image;
}

```

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,64 @@
...
+
+@RestController
+public class TestController {
+
+    @Autowired
+    private TestService testService;
+
+    @RequestMapping(value = "/tests/create",
+                    method = RequestMethod.POST,
+                    headers = "Content-Type=application/json;charset=utf-8")
```
stanislavgrujic: Please rather use value for `consumes` and `produces`.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @RequestMapping(value = "/tests/{userId}",
+                    method = RequestMethod.GET)
+    public ResponseEntity<List<TestResponse>> getTestsByUserId(@PathVariable("userId") Long userId) {
+        List<Test> tests = testService.findTestsByUserId(userId);
+        List<TestResponse> testResponses = tests.stream().map(TestResponse::new).collect(Collectors.toList());
+        return new ResponseEntity<>(testResponses, HttpStatus.OK);
+    }
+
+    @RequestMapping(value = "/deletetest/{testId}",
```
stanislavgrujic: Please map this method to `/test/{testId}` and `method = RequestMethod.DELETE`.

stanislavgrujic:
Please resolve merge conflicts.

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
stanislavgrujic: This comment is still not resolved.

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -22,10 +24,17 @@
...
   private TestRepository testRepository;
+  @Autowired
+  private UserRepository userRepository;
+  @Autowired
+  private FormulaRepository formulaRepository;
 
   @Override
   @Transactional
-  public Test create(Formula formula, User user) {
+  public Test create(Long formulaId, Long userId) {
```
stanislavgrujic: Why did you change the method signature? Having specific types in the method header instead of Long values give s you more type safety (e.g., with this method signature, you can easily switch `formulaId` and `userId` when calling the method, while it cannot happen when you use specific types).

web/src/main/java/eu/execom/testgenerator/web/dto/TestResponse.java
```diff
@@ -0,0 +1,29 @@
...
+    private Long id;
+    private boolean deleted;
+    private Long applicantId;
+    private Long reviewerId;
+    private Long formulaId;
+
+    public TestResponse(Test test) {
+        this.applicantName = test.getApplicant().getName();
+        this.questions = test.getTestQuestions().stream().map(TestQuestion::getQuestion).map(QuestionDto::new).collect(
+                Collectors.toList());
```
stanislavgrujic: How do you populate the other fields in this class with this constructor?

luinnarn: Is it better for the TestResponse object to hold a reference to the Test object as the constructor arg (private Test test ... this.test == test), or to populate all the fields manually in the constructor?

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -22,81 +22,83 @@
...
-  @Autowired
-  private UserRepository userRepository;
-  @Autowired
-  private FormulaRepository formulaRepository;
+    @Autowired
+    private TestRepository testRepository;
+    @Autowired
+    private UserRepository userRepository;
+    @Autowired
+    private FormulaRepository formulaRepository;
```
stanislavgrujic: The spacing should be 2 spaces for tab character. Is something changed in your formatter?

luinnarn: I suspect my formatter is misbehaving... I will check it out.

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -84,8 +86,24 @@ public Test findOne(Long id) {
 
   @Override
   @Transactional(readOnly = true)
-  public List<Test> findTestsByUserId(Long userId) {
-    return userRepository.getOne(userId).getTest();
+  public List<Test> findTestsByApplicantId(Long userId) {
+    User user = userRepository.getOne(userId);
+    if (user != null && user.getRole() == UserRole.APPLICANT) {
```
stanislavgrujic: Shouldn't call to `getOne()` already throw an `EntityNotFoundException`, if an entity is not found, is null check really needed?

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -84,8 +86,24 @@ public Test findOne(Long id) {
...
+      return userRepository.getOne(userId).getApplicantTests();
```
stanislavgrujic: Do you have to make another call to the `userRepository` to fetch user one more time, just in order to get his tests? Can't you just `return user.getApplicantTests()`?

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -30,20 +34,29 @@
...
+    if (userService.findByEmail(testRequest.getUserEmail()) != null) {
+      user = userService.findByEmail(testRequest.getUserEmail());
+    } else {
+      user = new User();
+      user.setRole(UserRole.APPLICANT);
+      user.setName(testRequest.getUserName());
+      user.setEmail(testRequest.getUserEmail());
+      userService.save(user);
+    }
+    Test test = testService.create(formulaService.findOne(testRequest.getFormulaId()), user);
```
stanislavgrujic: This logic should all go into service, not in the controller.

stanislavgrujic: Think of a controller just as a proxy to the service.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -54,29 +67,32 @@
...
+    List<Test> tests = new ArrayList<>();
     if (userId == 0) {
       tests = testService.findAll();
-    } else {
-      tests = testService.findTestsByUserId((long) userId);
+    } else if (userService.findOne((long) userId).getRole() == UserRole.APPLICANT) {
+      tests = testService.findTestsByApplicantId((long) userId);
+    } else if (userService.findOne((long) userId).getRole() == UserRole.REVIEWER) {
+      tests = testService.findTestsByReviewerId((long) userId);
     }
```
stanislavgrujic: Please move logic to the service

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
stanislavgrujic: Once you move the logic to the service, please add unit tests for it.

core/src/main/java/eu/execom/testgenerator/core/model/User.java
```diff
@@ -51,6 +51,23 @@
...
+  @OneToMany(mappedBy = "applicant",
+             cascade = CascadeType.ALL)
+  private List<Test> applicantTests;
+
+  @OneToMany(mappedBy = "reviewer",
+             cascade = CascadeType.ALL)
+  private List<Test> reviewerTests;
+
+  public boolean isApplicant() {
+    return (this.getRole() == UserRole.APPLICANT ? true : false);
```
stanislavgrujic: Please rewrite the method body as (and the ones below):
```
return this.role == UserRole.APPLICANT;
```

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -75,6 +84,26 @@ public Test findOne(Long id) {
...
+      User user = userRepository.getOne(userId);
+      if (user.isReviewer()) {
+        return user.getReviewerTests();
+      } else if (user.isApplicant()) {
+        return user.getApplicantTests();
+      } else {
+        return findAll();
+      }
+    } else {
+      throw new EntityNotFoundException();
```
stanislavgrujic: This branch is probably never reachable since `userRepository.getOne()` will throw the exception if the user is not found.

core/src/main/java/eu/execom/testgenerator/core/service/UserServiceImpl.java
```diff
@@ -20,25 +21,47 @@ public User save(User user) {
...
     userRepository.delete(id);
   }
+
+  @Override
+  @Transactional
+  public User getUserByNameAndEmail(String name, String email) {
+    User user;
+    if (findByEmail(email) != null) {
+      user = findByEmail(email);
+    } else {
```
stanislavgrujic: This is very inefficient - you will call `findByEmail()` method two times.
Please rewrite as
```
User user = findByEmail(email);
if (user == null) {
  // the rest of logic
}
```

core/src/main/java/eu/execom/testgenerator/core/service/UserServiceImpl.java
```diff
@@ -20,25 +21,47 @@ public User save(User user) {
...
+      user = new User();
+      user.setRole(UserRole.APPLICANT);
+      user.setName(name);
+      user.setEmail(email);
+      save(user);
```
stanislavgrujic: Are we sure that we want the method getByEmail to actually create a new user?

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,79 @@
...
+  @Autowired
+  private UserService userService;
+
+  @RequestMapping(value = "/create",
+                  method = RequestMethod.POST,
+                  consumes = MediaType.APPLICATION_JSON_VALUE,
+                  produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<TestResponse> createTest(
+      @RequestBody
+          TestRequest testRequest) {
```
stanislavgrujic: Please inline these parameters.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,79 @@
...
+    TestResponse response = new TestResponse(test);
+    return new ResponseEntity<>(response, HttpStatus.OK);
+  }
+
+  @RequestMapping(value = "/{testId}",
+                  method = RequestMethod.GET,
+                  produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<TestResponse> getTestById(
+      @PathVariable("testId")
+          Long testId) {
```
stanislavgrujic: Please inline these parameters with the method header.

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
luinnarn: This branch should cover cases where a nonexistent id is passed as `RequestParam`, in order to avoid returning all tests in such cases.

core/src/main/java/eu/execom/testgenerator/core/service/UserServiceImpl.java
luinnarn: The method `getUserByNameAndEmail` is being called during test creation, at which moment a new user should be created if not found in the database, if I understood the wiki correctly... What's missing at this point is functionality to check if there is an existing user with the same full name, in which case the system asks if the user has already taken the test, and asks for the e-mail which was used previously

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
luinnarn: That darn formatter is at it again

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
luinnarn: ... if I'm not mistaken. I'll try to find a better solution.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,71 @@
...
+    TestResponse response = new TestResponse(test);
+    return new ResponseEntity<>(response, HttpStatus.CREATED);
+  }
+
+  @RequestMapping(method = RequestMethod.GET,
+                  produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<List<TestResponse>> getTests(@RequestParam(name = "userId",
+                                                                   defaultValue = "0") int userId) {
+    List<Test> tests;
+    tests = testService.findTestsByUserId((long) userId);
```
stanislavgrujic: Please do not split the variable declaration and definition unless there is really need to do that -> make this two lines a single one.

web/src/main/java/eu/execom/testgenerator/web/dto/TestResponse.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+@NoArgsConstructor
+@Data
+public class TestResponse {
+
+  private Long id;
+  private boolean deleted;
```
stanislavgrujic: This field should not exist in the response. If a test is deleted, it should not be retrieved from the database and the UI should not know about it.

web/src/main/java/eu/execom/testgenerator/web/dto/TestResponse.java
```diff
@@ -0,0 +1,38 @@
...
+  private Long formulaId;
+
+  private List<QuestionDto> questions;
+
+  public TestResponse(Test test) {
+    if (test != null) {
+      this.id = test.getId();
+      this.deleted = test.isDeleted();
+      this.applicantId = test.getApplicant().getId();
+      this.reviewerId = ((test.getReviewer() == null)? null : test.getReviewer().getId());
```
stanislavgrujic: Please fix formatting.

web/src/test/java/eu/execom/testgenerator/web/controller/TestControllerTest.java
```diff
@@ -0,0 +1,139 @@
...
+import static org.mockito.Mockito.when;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(MockitoJUnitRunner.class)
+@TestExecutionListeners({})
```
stanislavgrujic: Please remove this annotation, it's not used.

web/src/test/java/eu/execom/testgenerator/web/controller/TestControllerTest.java
```diff
@@ -0,0 +1,139 @@
...
+    mockMvc = MockMvcBuilders.standaloneSetup(testController).addFilters(new CorsFilter()).build();
+  }
+
+  @org.junit.Test
+  public void shouldGetAllTests() throws Exception {
+    User applicant = applicant(2L);
+    Formula formula = formula("FormulaName", 3L);
+    Test test1 = test(1L, formula, applicant, new ArrayList<TestQuestion>());
+    Test test2 = test(2L, formula, applicant, new ArrayList<TestQuestion>());
+    Test test3 = test(3L, formula, applicant, new ArrayList<TestQuestion>());
```
stanislavgrujic: If you're always passing an empty ArrayList you can also internalize it in the method body.

web/src/test/java/eu/execom/testgenerator/web/controller/TestControllerTest.java
```diff
@@ -0,0 +1,139 @@
...
+    User user = new User();
+    user.setId(3L);
+
+    //when
+    when(formulaService.findOne(1L)).thenReturn(formula);
+    when(userService.getUserByNameAndEmail(anyString(), anyString())).thenReturn(user);
+
+    //then
+    mockMvc.perform(post(URL_PREFIX + "/create").content(
+        "{\"formulaId\" : 1,\"userName\" : \"Slavko Stimac\",\"userEmail\" : \"salas@umalomritu\"}")
```
stanislavgrujic: If we add e-mail validatiom, this request will fail it. Please use proper email addresses.

web/src/main/java/eu/execom/testgenerator/web/dto/TestRequest.java
```diff
@@ -0,0 +1,12 @@
...
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
+@Data
+public class TestRequest {
+  private Long formulaId;
+  private String userName;
+  private String userEmail;
```
stanislavgrujic: Please add request validation.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,71 @@
...
+  @Autowired
+  private FormulaService formulaService;
+  @Autowired
+  private UserService userService;
+
+  @RequestMapping(value = "/create",
+                  method = RequestMethod.POST,
+                  consumes = MediaType.APPLICATION_JSON_VALUE,
+                  produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<TestResponse> createTest(@RequestBody TestRequest testRequest) {
```
stanislavgrujic: Please add @Valid annotation for the request body, and also add validation annotations to the `TestRequest`.

core/src/test/java/eu/execom/testgenerator/core/service/TestServiceTest.java
```diff
@@ -54,10 +73,23 @@ public void setUp() {
...
+      @Override
+      public List<Test> answer(InvocationOnMock invocationOnMock) throws Throwable {
+        return allTests;
+      }
+    });
+    given(userRepository.getOne(any(Long.TYPE))).willAnswer(new Answer<User>() {
+      @Override
+      public User answer(InvocationOnMock invocationOnMock) throws Throwable {
+        return user;
+      }
```
stanislavgrujic: Please use lambda expressions and Java8 features to define the mocks behavior.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -0,0 +1,75 @@
...
+public class TestController {
+
+  @Autowired
+  private TestService testService;
+  @Autowired
+  private FormulaService formulaService;
+  @Autowired
+  private UserService userService;
+
+  private final Logger logger = LoggerFactory.getLogger(TestController.class);
```
stanislavgrujic: Please use Lombok annotation `@Slf4j`.

web/src/test/java/eu/execom/testgenerator/web/controller/TestControllerTest.java
```diff
@@ -0,0 +1,133 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(MockitoJUnitRunner.class)
+public class TestControllerTest {
+
+  private static final String URL_PREFIX = "/tests";
+
+  private static final MediaType JSON_CONTENT_TYPE = new MediaType(MediaType.APPLICATION_JSON.getType(),
+      MediaType.APPLICATION_JSON.getSubtype(), Charset.forName("utf8"));
```
stanislavgrujic: Can't you just use `MediaType.APPLICATION_JSON` in code, instead of having to create this constant?

PR #33, opened 2017-12-14
Add web application exception handler

### LINKED ISSUE #25: Add a web application exception handler
Based on the caught exception return one of the following status codes:

1. Validation of input parameters fails - 400
2. User is not authenticated in the system - 401
3. User does not have rights to access a specific resource - 403
4. Entity is not found in the repository by an id - 404
5. Any other exception - 500

nikbas:
Closes #25


nikbas:
Not sure which exception classes should be used.

web/src/main/java/eu/execom/testgenerator/web/util/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+  @ExceptionHandler(InsufficientScopeException.class)
+  public void handleInsufficientScopeException() {}
+
+  @ResponseStatus(HttpStatus.NOT_FOUND) // 404
+  @ExceptionHandler(EntityNotFoundException.class)
+  public void handleEntityNotFoundException() {}
+
+  @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR) // 500
+  @ExceptionHandler(Exception.class)
+  public void handleException() {}
```
stanislavgrujic: Please add exception logging on all of these methods instead of empty method bodies.

stanislavgrujic:
Please have a look at the `FormulaControllerTest` and update it accordingly.

stanislavgrujic:
Now that you have FormulaController you can easily check which exceptions are being thrown in case that request body validation fails, an entity is not found and you could cover other that you can find there as well. :)

web/src/main/java/eu/execom/testgenerator/web/util/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,61 @@
...
+  public void handleEntityExistsException(HttpServletRequest request, Exception exception) {
+
+    log.error("Requested URL: {}. Exception raised: {}", request.getRequestURI(), exception.toString());
+  }
+
+  @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR) // 500
+  @ExceptionHandler(Exception.class)
+  public void handleException(HttpServletRequest request, Exception exception) {
+
+    log.error("Requested URL: {}. Exception raised: {}", request.getRequestURI(), exception.toString());
```
stanislavgrujic: Two remarks on this -
1) no need to explicitly call `toString()`, log4j will call it (similar as when doing the string concatenation).
2) Please extract this line to a method, e.g.,
```
private void logError(HttpServletRequest request, Exception exception) {
  log.error("Requested URL: {}. Exception raised: {}", request.getRequestURI(), exception.toString());
}
```

and reuse it.

stanislavgrujic:
Could you also update the existing controller tests to reflect the updated code?

PR #43, opened 2018-01-03
Controller integration tests

luinnarn:


pom.xml
```diff
@@ -101,6 +101,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
```
stanislavgrujic: Please use the `test` scope.

pom.xml
```diff
@@ -101,6 +101,12 @@
...
+    </dependency>
+
+
```
stanislavgrujic: Please remove useless blank lines.

web/src/main/java/eu/execom/testgenerator/web/controller/CategoryController.java
```diff
@@ -27,14 +29,19 @@
...
                   consumes = MediaType.APPLICATION_JSON_VALUE,
                   produces = MediaType.APPLICATION_JSON_VALUE)
-  public ResponseEntity<CategoryResponse> createNewCategory(@RequestBody @Valid CreateCategoryRequest createCategoryRequest) {
+  public ResponseEntity<CategoryResponse> createNewCategory(
+      @RequestBody @Valid CreateCategoryRequest createCategoryRequest) {
     Category category = convertCreateCategoryRequestToCategory(createCategoryRequest);
-    category = categoryService.save(category);
+    try {
+      category = categoryService.save(category);
+    } catch (EntityExistsException exception) {
```
stanislavgrujic: This should be, and probably already is, handled in the `GlobalExceptionHandler`.

stanislavgrujic: Please remove all try/catch where you added such logic.

web/src/main/java/eu/execom/testgenerator/web/dto/QuestionResponse.java
```diff
@@ -37,6 +36,8 @@ public QuestionResponse(Question question) {
     this.type = question.getType();
     this.category = question.getCategory().getId();
     this.codeImage = question.getCodeImage();
-    question.getAnswers().forEach(answer -> answers.add(new AnswerResponse(answer)));
+    if (question.getAnswers() != null) {
```
stanislavgrujic: Rather make sure that answers are never null.

web/src/main/java/eu/execom/testgenerator/web/controller/QuestionController.java
```diff
@@ -55,23 +56,31 @@
...
   @RequestMapping(method = RequestMethod.PUT,
                   consumes = MediaType.APPLICATION_JSON_VALUE,
                   produces = MediaType.APPLICATION_JSON_VALUE)
   public ResponseEntity<QuestionResponse> updateQuestion(@RequestBody @Valid UpdateQuestionRequest updateQuestionRequest) {
     Question question = convertUpdateQuestionRequestToQuestion(updateQuestionRequest);
-    question = questionService.update(question);
+    try {
+      question = questionService.update(question);
+    } catch (EntityNotFoundException exception) {
+      return new ResponseEntity<>(HttpStatus.NOT_FOUND);
```
stanislavgrujic: Please remove all try/catch clauses which you added. Exception handling logic should be in the `GlobalExceptionHandler`.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -59,6 +59,9 @@
                   produces = MediaType.APPLICATION_JSON_VALUE)
   public ResponseEntity<TestResponse> getTestById(@PathVariable("testId") Long testId) {
     Test test = testService.findOne(testId);
+    if (test == null) {
+      return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+    }
```
stanislavgrujic: If this is needed, rather update `testService.findOne(testId)`, to use `testRepository.getOne(testId)`, in order to throw an exception, which will then be handled in the `GlobalExceptionHandler`. Also, please make sure that `findOne` method is consistently implemented in all service methods.

web/src/test/java/eu/execom/testgenerator/web/QuestionMother.java
```diff
@@ -12,6 +12,76 @@ private QuestionMother() {
...
+  }
+
+  public static Question question7(Category category) {
+    Question question = new Question();
+    question.setText("Question text 7");
+    question.setType(QuestionType.FREE_FORM);
+    question.setPoints(18);
+    question.setCategory(category);
+    question.setId(7L);
+    return question;
```
stanislavgrujic: Instead of duplicating code for creation of a free form question, maybe you could make one method like this
```
public static Question freeFormQuestion(Long id, Category category, int points) {
}
```
and reuse `id` for `text` and `id`.

web/src/test/java/eu/execom/testgenerator/web/RecipeMother.java
```diff
@@ -0,0 +1,27 @@
...
+    return recipe;
+  }
+
+  public static Recipe recipe2(Category category) {
+    Recipe recipe = new Recipe();
+    recipe.setCategory(category);
+    recipe.setNumberOfQuestions(3);
+    recipe.setId(2L);
+    return recipe;
+  }
```
stanislavgrujic: Please see my comment above for reusal of code.

web/src/test/java/eu/execom/testgenerator/web/controller/CategoryControllerTest.java
```diff
@@ -27,129 +23,97 @@
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
+@RunWith(SpringRunner.class)
+@SpringBootTest
```
stanislavgrujic: The idea was not to replace unit tests with integration tests, but to add integration tests alongside the unit tests.

stanislavgrujic: Please revert back to unit tests, and add a new file with this logic.

web/src/test/java/eu/execom/testgenerator/web/controller/DBPersistenceTest.java
```diff
@@ -0,0 +1,30 @@
...
+  private TestRepository testRepository;
+
+  @org.junit.Test
+  @Transactional
+  public void shouldSaveAndRetrieveTest() {
+    Test savedTest = testRepository.save(new Test());
+    Test retreivedTest = testRepository.findOne(savedTest.getId());
+
+    assertNotNull(retreivedTest);
+    assertEquals(retreivedTest, savedTest);
```
stanislavgrujic: This is not needed to test, since we don't implement repositories ourselves. Please remove it.

core/src/main/java/eu/execom/testgenerator/core/model/User.java
```diff
@@ -70,4 +71,27 @@ public boolean isReviewer() {
...
+
+  public void addTest (Test test) {
+    switch (role) {
+      case APPLICANT: {
+        if (applicantTests == null) {
+          applicantTests = new ArrayList<>();
+        }
+        applicantTests.add(test);
+        break;
+      }
```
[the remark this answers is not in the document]
luinnarn: I've encountered multiple places where null checks turned out to be necessary. Would it perhaps be suitable to use the `Optional` interface?

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+public class CategoryControllerIntegrationTest {
+
+  private static final String URL_PREFIX = "/categories";
+  private static final MediaType contentType = MediaType.APPLICATION_JSON_UTF8;
```
stanislavgrujic: Please remove this constant and use the `MediaType.APPLICATION_JSON_UTF8` instead on place where it is needed.

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Test
+  @Transactional
+  public void shouldCreateNewCategory() throws Exception {
+    String newCategory = "{\"name\" : \"A category\"}";
+
+    mockMvc.perform(post(URL_PREFIX).content(newCategory).contentType(contentType))
+           .andExpect(status().isCreated())
+           .andExpect(content().contentType(contentType));
```
stanislavgrujic: How do you know that the category is created? You should invoke another operation via controller to fetch it, in order to make sure that it is created.

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Before
+  public void init() {
+    mockMvc = MockMvcBuilders.standaloneSetup(categoryController)
+                             .setControllerAdvice(new GlobalExceptionHandler())
+                             .build();
+  }
+
+  @Test
+  @Transactional
```
stanislavgrujic: Why do the tests need this?

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    mockMvc.perform(post(URL_PREFIX).content(newCategory).contentType(contentType))
+           .andExpect(status().isCreated())
+           .andExpect(content().contentType(contentType));
+  }
+
+  @Test
+  @Transactional
+  public void shouldReturnConflictOnCreateNewCategoryWhenNameAlreadyExists() throws Exception {
+    Category category = categoryService.save(category(1L,"Same name"));
```
stanislavgrujic: Would be better if you could instead of invoking methods on category service do all actions that affect the database state through the controller itself.

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+    mockMvc.perform(get(URL_PREFIX))
+           .andExpect(status().isOk())
+           .andExpect(content().contentType(contentType))
+           .andExpect(jsonPath("$", hasSize(3)));
+  }
+
+  @Test
+  @Transactional
+  public void shouldGetCategoryByName() throws Exception {
+    Category category1 = categoryService.save(category(1L,"First category"));
```
stanislavgrujic: Please pay attention to formatting and IntellijI IDEA's inspections that point you to this kind of things.

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    mockMvc.perform(get(URL_PREFIX + "/First category"))
```
stanislavgrujic: Are you sure that browsers handle white space this way? You should try to make it as similar to the real world as possible.

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Test
+  @Transactional
+  public void shouldUpdateCategory() throws Exception {
+    Category category1 = categoryService.save(category(null,"Old name"));
+    String categoryJSON = "{\"name\":\"New name\", \"id\":\"" + category1.getId() + "\"}";
+
+    mockMvc.perform(put(URL_PREFIX).contentType(contentType).content(categoryJSON))
+           .andExpect(status().isOk())
+           .andExpect(content().contentType(contentType));
```
stanislavgrujic: You should create another call to the controller to fetch this category to make sure that it is updated properly in the database.

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Test
+  @Transactional
+  public void shouldReturnIsNotFoundOnUpdateCategoryIdDoesNotExist() throws Exception {
+    String categoryJSON = "{\"name\":\"New name\", \"id\":\"9\"}";
```
stanislavgrujic: Please extract this as a constant. Also, camel case naming convention suggests that the abbreviations should also be written only with first letter in uppercase (e.g.,`Json` instead of `JSON`).

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    mockMvc.perform(put(URL_PREFIX).contentType(contentType).content(categoryJSON)).andExpect(status().isNotFound());
+  }
+
+  @Test
+  @Transactional
+  public void shouldDeleteCategory() throws Exception {
+    Category category1 = categoryService.save(category(null,"First category"));
+
+    mockMvc.perform(delete(URL_PREFIX + "/" + category1.getId())).andExpect(status().isNoContent());
```
stanislavgrujic: You should make another call through the controller to make sure that the category is actually deleted (i.e., it does not exist anymore)

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
luinnarn: So they would not affect each other in terms of persisting to the database (to roll back after each test)

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @Test
+  @Transactional
+  public void shouldUpdateCategory() throws Exception {
+
+    mockMvc.perform(
+        post(URL_PREFIX).content(categoryRequestJson(OLD_NAME)).contentType(MediaType.APPLICATION_JSON_UTF8));
+    String categoryUpdateJson =
+        "{\"name\":\"New name\", \"id\":\"" + categoryService.findByName(OLD_NAME).getId() + "\"}";
```
stanislavgrujic: It is a bad practice to include method calls in string concatenation, since it is very difficult to spot it. Please extract it to a variable.

web/src/test/java/eu/execom/testgenerator/web/integration/CategoryControllerIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+  @Test
+  @Transactional
+  public void shouldDeleteCategory() throws Exception {
+
+    mockMvc.perform(
+        post(URL_PREFIX).content(categoryRequestJson(OLD_NAME)).contentType(MediaType.APPLICATION_JSON_UTF8))
+           .andExpect(status().isCreated())
+           .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8));
+
+    mockMvc.perform(delete(URL_PREFIX + "/" + categoryService.findByName(OLD_NAME).getId()))
```
stanislavgrujic: Please extract service call to a variable.

PR #49, opened 2018-02-02
add mobx routing animations login-page

aberonja:
Add mobx, routing, animations, login-page, images that are necessary.


web/src/main/frontend/src/scss/components/_animation.scss
```diff
@@ -0,0 +1,31 @@
+.SlideIn-appear {
+  transform: translateY(-30px);
```
iscekic: Should `30px` be a variable?

web/src/main/frontend/src/scss/layout/_startLayout.scss
```diff
@@ -0,0 +1,47 @@
+.content__block {
+  background-image: url('img/background.png');
+  background-size: cover;
+  height: 100vh;
+  background-repeat: no-repeat;
+  background-position-x: center;
+  .login__block {
+    padding: 50px 5%;
```
iscekic: Avoid using percent values for properties other than `height` and `width`.

web/src/main/frontend/src/App.js
```diff
@@ -1,11 +1,26 @@
...
+          <Route path={'login'} component={StartScreen}>
+            <IndexRoute component={Login} />
+          </Route>
+        </Route>
+      </Router>
+    );
   }
 }
 
+App = inject('Store')(observer(App));
```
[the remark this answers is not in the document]
iscekic: Why is this component decorated in the first place? It isn't displaying anything from the `Store`.

web/src/main/frontend/src/index.js
```diff
@@ -2,6 +2,18 @@ import React from 'react';
...
 import App from './App';
 import registerServiceWorker from './registerServiceWorker';
+import Store from './base/Store';
+import {Provider} from 'mobx-react';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+class Root extends React.Component {
+  render() {
+    return (
+      <Provider Store={Store}>
```
iscekic: Prop names should be lowercase, eg. `store={Store}`

web/src/main/frontend/src/layout/StartScreen.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import execomLogo from '../img/logo-login.png';
+
+export class StartScreen extends React.Component {
```
iscekic: No need to export this class.

web/src/main/frontend/src/pages/Login.js
```diff
@@ -0,0 +1,27 @@
...
+          <h3 className="text-left">Login</h3>
+          <Link to={'/'}>I want to be Execom employee</Link>
+          <Link to={'/'}>I am Execom employee</Link>
+        </div>
+      </ReactCSSTransitionGroup>
+    );
+  }
+}
+
+Login = inject('Store')(observer(Login));
```
iscekic: Same comment as above, don't overwrite variables, why is it decorated.

web/src/main/frontend/src/pages/Login.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import {observer, inject} from 'mobx-react';
+import {Link} from 'react-router';
+import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+
+export class Login extends React.Component {
```
iscekic: No need to export this class.

web/src/main/frontend/src/scss/components/_animation.scss
iscekic: `0.6` too, maybe (your call).

web/src/main/frontend/src/scss/layout/_startLayout.scss
```diff
@@ -0,0 +1,47 @@
+.content__block {
```
iscekic: If any of the numbers in this file is used more than once, please extract them to a variable.

web/src/main/frontend/src/scss/layout/_startLayout.scss
```diff
@@ -0,0 +1,47 @@
...
+        margin-bottom: 150px;
+      }
+    }
+    h3 {
+      margin-bottom: 60px;
+      color: white;
+      font-size: 35px;
+    }
+    .google__button,
+    a {
```
iscekic: Will this class be used in parents which won't be `component__block`s? If yes, extract it.

web/src/main/frontend/src/pages/Login.js
```diff
@@ -0,0 +1,27 @@
...
+  render() {
+    return (
+      <ReactCSSTransitionGroup
+        transitionName="SlideIn"
+        transitionAppear={true}
+        transitionAppearTimeout={600}
```
iscekic: Can these numbers be extracted to a `const` variable?

web/src/main/frontend/src/pages/Login.js
```diff
@@ -0,0 +1,27 @@
...
+import {observer, inject} from 'mobx-react';
+import {Link} from 'react-router';
+import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+
+export class Login extends React.Component {
+  render() {
+    return (
+      <ReactCSSTransitionGroup
+        transitionName="SlideIn"
+        transitionAppear={true}
```
iscekic: No need to explicitly set true, just `transitionAppear` is enough.

web/src/main/frontend/src/pages/Login.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import {observer, inject} from 'mobx-react';
+import {Link} from 'react-router';
+import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+
+export class Login extends React.Component {
+  render() {
+    return (
+      <ReactCSSTransitionGroup
+        transitionName="SlideIn"
```
iscekic: Transition name should be `slide-in`.

web/src/main/frontend/src/scss/base/_base.scss
```diff
@@ -17,13 +17,13 @@ a {
   &:active {
     text-decoration: none;
     outline: none;
-    @include transition-mix(0.2s);
+    @include ease-in-out(0.2s);
```
iscekic: `0.2` to variable?

web/src/main/frontend/src/scss/components/_animation.scss
iscekic: There is lots of duplicate code in this file, find a way to DRY-ify it.

web/src/main/frontend/src/App.js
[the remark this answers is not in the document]
iscekic: Actually, for future cases, we can make the export and injection a one-liner, probably something like:
```js
export default inject('store')(observer(App))
```

web/src/main/frontend/src/scss/components/_animation.scss
aberonja: I wouldn't make `30px` a variable cause it's only used on two places (after DRY-ifing), and I'm still not sure that animation will look like this. If you think it's better to make it var I'll do it.

web/src/main/frontend/src/pages/Login.js
```diff
@@ -1,27 +1,30 @@
...
+        transitionAppearTimeout={durationTime}
+        transitionEnterTimeout={durationTime}
+        transitionLeaveTimeout={durationTime}
       >
         <div>
           <h3 className="text-left">Login</h3>
-          <Link to={'/'}>I want to be Execom employee</Link>
-          <Link to={'/'}>I am Execom employee</Link>
+          <Link className="button__link" to="/">
+            I want to be Execom employee
```
iscekic: The sentences should be:
`I want to be an Execom employee`
`I am an Execom employee`

web/src/main/frontend/src/App.js
```diff
@@ -1,10 +1,15 @@
...
 import './App.css';
 import Root from './components/Root';
+import {BrowserRouter} from 'react-router-dom';
 
 class App extends Component {
   render() {
-    return <Root />;
+    return (
+      <BrowserRouter>
+        <Root />
```
iscekic: Can we completely avoid having a `Root` component by having the router config here?

web/src/main/frontend/src/scss/components/_animation.scss
```diff
@@ -0,0 +1,26 @@
+.slide-in {
+  &-appear,
+  &-enter {
+    transform: translateY(-30px);
```
iscekic: `30px` to variable?

web/src/main/frontend/src/pages/Login.js
```diff
@@ -0,0 +1,30 @@
...
+    return (
+      <ReactCSSTransitionGroup
+        transitionName="slide-in"
+        transitionAppear
+        transitionAppearTimeout={durationTime}
+        transitionEnterTimeout={durationTime}
+        transitionLeaveTimeout={durationTime}
+      >
+        <div>
+          <h3 className="text-left">Login</h3>
```
iscekic: This should probably be `h1`.

web/src/main/frontend/src/scss/layout/_startLayout.scss
```diff
@@ -0,0 +1,47 @@
...
+    }
+    @media (min-width: 1199px) {
+      padding: 100px 130px;
+    }
+    background-color: $gray-color;
+    img {
+      width: 70%;
+      margin-bottom: 50px;
+      @media (min-width: 545px) {
+        margin-bottom: 150px;
```
iscekic: This seems wrong to me. Please check this file for numbers which should be variables (variables can be local to this file, they needn't be in `_variables.scss`).

Once you check that, I'll come to you so we can see if there's a better way to accomplish the same thing.

web/src/main/frontend/package.json
```diff
@@ -11,8 +11,10 @@
     "npm-run-all": "^4.1.2",
     "prettier": "^1.10.2",
     "react": "^16.2.0",
+    "react-addons-css-transition-group": "^15.6.2",
     "react-dom": "^16.2.0",
-    "react-router": "^4.2.0",
+    "react-google-login": "^3.0.4",
```
iscekic: Is this library used as part of this PR? If not, please remove it.

web/src/main/frontend/src/App.js
iscekic: We have `index`, `App` and `Root`. We should only have `index` and `App`.

web/src/main/frontend/src/scss/components/_animation.scss
aberonja: I thought about removing vertical movement of animation, and only leave fade-in-out effect. Tell me what do you think about that.

iscekic: I don't really have a strong opinion. Consult with the designer. :smile:

PR #54, opened 2018-02-07
add select on login page and Services.js file

aberonja:
Created Select.js component and appropriate style in _login.scss file
Removed export from GuestLogin.js and added Select component
Created Services.js file for functions
Added color change function for select component in Services.js file
Fixed responsive issues on start screen


web/src/main/frontend/src/base/Services.js
```diff
@@ -0,0 +1,3 @@
+export const selectChangeColor = () => {
+  document.getElementById('test-type').style.color = 'white';
```
iscekic: `getElementById`? :no_good_man:

web/src/main/frontend/src/pages/GuestLogin.js
```diff
@@ -36,12 +37,18 @@ export class GuestLogin extends React.Component {
                 placeholder="Please enter your email"
               />
             </div>
+            <div className="col-12 col-md-3 text-left">
+              <label htmlFor="testType">Test type</label>
```
iscekic: IDs should use hyphens.

web/src/main/frontend/src/base/Services.js
aberonja: What alternative do you suggest? :smile:

iscekic: Local component state.

web/src/main/frontend/src/components/Select.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+class Select extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      change: false,
+      changeFunc: function handleChange() {
```
iscekic: Functions shouldn't be part of `state`. Use a method:
```js
class ... extends ... {
  changeColor = () => {
    this.setState({
      color: 'new color'
    });
  }
}
```

web/src/main/frontend/src/components/Select.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+class Select extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
iscekic: You can avoid writing a constructor by using this syntax:
```js
class ... extends ... {
  state = {
    color: 'initial color'
  }
}
```

web/src/main/frontend/src/components/Select.js
```diff
@@ -0,0 +1,39 @@
...
+      change: false,
+      changeFunc: function handleChange() {
+        this.setState({change: true});
+      }
+    };
+    this.state.changeFunc = this.state.changeFunc.bind(this);
+  }
+
+  render() {
+    const selectStyle = {
```
iscekic: Remove this variable.

web/src/main/frontend/src/components/Select.js
```diff
@@ -0,0 +1,39 @@
...
+      color: 'white'
+    };
+    return (
+      <div className="col-12 col-md-9 text-left">
+        <select
+          id="test-type"
+          style={this.state.change ? selectStyle : null}
```
iscekic: This line should be:
```js
style={{
  color: this.state.color
}}
```

web/src/main/frontend/src/components/Select.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+class Select extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      change: false,
+      changeFunc: function handleChange() {
+        this.setState({change: true});
```
iscekic: You should never `setState` in the constructor.

web/src/main/frontend/src/components/Select.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+class Select extends React.Component {
+  state = {
+    color: null
```
iscekic: `null` is not a valid initial color. :smile: Or does React handle that case by not inserting CSS?

aberonja: React handles it. There is no style on this element before change.

PR #61, opened 2018-02-13
Add admin panel routing

aberonja:
In App.js changed routes
Added components for admin panel
Added Panel.js component for admin panel
Added _panel.scss file for admin panel styling
Changed favicon


web/src/main/frontend/src/components/Category.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+
+class Category extends React.Component {
+  render() {
+    const durationTime = 600;
```
iscekic: Define this variable outside `render`.

web/src/main/frontend/src/pages/Panel.js
```diff
@@ -0,0 +1,89 @@
...
+                <div className="nav-link__block">
+                  <a href="/login">
+                    <img src={LoginIcon} alt="login icon" />
+                    <span>Login page</span>
+                  </a>
+                </div>
+              </nav>
+            </div>
+            <Switch>
+              <Route exact path={`/dashboard`} component={Dashboard} />
```
iscekic: Is `exact` needed for all of these?

web/src/main/frontend/src/pages/Panel.js
```diff
@@ -0,0 +1,89 @@
...
+              </div>
+              <nav>
+                <div className="nav-link__block">
+                  <NavLink exact to="/dashboard" activeClassName="active">
+                    <img src={DashboardIcon} alt="dashboard icon" />
+                    <span>Dashboard</span>
+                  </NavLink>
+                </div>
+                <div className="nav-link__block">
+                  <NavLink to="/category" activeClassName="active">
```
iscekic: Extract this component and it's contents into a separate component.

web/src/main/frontend/src/layout/StartScreen.js
```diff
@@ -8,34 +8,35 @@ import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
...
+                <img src={execomLogo} alt="execom logo" />
               </div>
+              <Switch>
+                <Route
+                  exact
+                  path={`${this.props.match.path}`}
+                  component={Login}
+                />
+                <Route
+                  exact
```
iscekic: Is exact needed here?

web/src/main/frontend/src/components/Category.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+
+class Category extends React.Component {
```
iscekic: `Category`, `Dashboard`, `Formula`, `Question` and `Review` are all the same. Use a single component for all of them.

web/src/main/frontend/src/pages/Panel.js
aberonja: Tried it now without exact and it works.

web/src/main/frontend/src/layout/StartScreen.js
aberonja: In StartScreen.js exact is needed because path is defined like this:
*path={`${this.props.match.path}/guest`}*
If I define it like this:
*path={`/login/guest`}*
then it works without exact.
What should I use? :smile:

web/src/main/frontend/src/components/Category.js
aberonja: Since it's used on a lot of pages can I define it in *store.js*?

iscekic: No, it should live in `env`.

web/src/main/frontend/src/layout/StartScreen.js
iscekic: Leave it as it is then. :smile:

web/src/main/frontend/src/pages/Panel.js
```diff
@@ -0,0 +1,95 @@
...
+        transitionLeaveTimeout={animationDurationTime}
+      >
+        <div className="container-fluid h-100 dashboard__block d-flex align-items-center justify-content-center">
+          <div className="row w-100 justify-content-center">
+            <div className="col-12 col-lg-3 col-xl-2 dashboard-side">
+              <div className="text-center logo__block">
+                <img src={execomLogo} alt="execom logo" />
+              </div>
+              <nav>
+                <NavigationLink
```
iscekic: Make an array called `routes`, which contains all the routes (obviously :smile:). Iterate through that array instead of hand-writing `NavigationLink`. Do the same in router `Switch` below.

web/src/main/frontend/src/pages/Panel.js
```diff
@@ -0,0 +1,95 @@
...
+                  name="Login"
+                  url={'/login'}
+                  image={`${LoginIcon}`}
+                />
+              </nav>
+            </div>
+            <Switch>
+              <Route
+                path={`${this.props.match.path}dashboard`}
+                render={props => <PanelContainer {...props} name="Dashboard" />}
```
iscekic: Only pass `PanelContainer` the props it needs.

web/src/main/frontend/src/pages/Panel.js
```diff
@@ -14,6 +14,29 @@ import PanelContainer from '../components/PanelContainer';
 
 class Panel extends React.Component {
   render() {
+    const routes = [
```
iscekic: Routes should be outside `render`. Maybe as a class property?

```js
class Panel extends React.Component {
  routes = [...];

  render() {...}
}
```

web/src/main/frontend/src/pages/Panel.js
```diff
@@ -14,6 +14,29 @@ import PanelContainer from '../components/PanelContainer';
...
+      {name: 'Dashboard', url: '/dashboard', image: {name: DashboardIcon}},
+      {name: 'Category', url: '/category', image: {name: CategoryIcon}},
+      {name: 'Question', url: '/question', image: {name: QuestionIcon}},
+      {name: 'Formula', url: '/formula', image: {name: FormulaIcon}},
+      {name: 'Review test page', url: '/review', image: {name: ReviewIcon}},
+      {name: 'Login page', url: '/login', image: {name: LoginIcon}}
+    ];
+    const routeList = routes.map(routes => (
```
iscekic: `routeList` and `switchList` shouldn't be variables. I think it would be better to inline them inside `render`:

```js
<nav>
  {this.routes.map(routes => (...)}
</nav>
```

iscekic: Don't forget that elements produced by `map` need to have a `key` prop.

aberonja: Is it okay to have key prop like this `key={routes.name}`?
I saw it on `reactjs.org`

iscekic: That's ok, as long as names will remain unique.

PR #67, opened 2018-02-20
Create Category and Formula components

aberonja:
Created Category.js component
Created Formula.js component
In PanelContainer.js added routing for those components
Created _category.scss and _formula.scss
In main.scss added _category.scss and _formula.scss
In _login.scss added style for inverted label and input
In _panel.scss added style for panel content
In _startLayout.scss added style for inverted button__link


web/src/main/frontend/src/components/Category.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+
+class Category extends React.Component {
+  render() {
+    return (
+      <div className="row dashboard__row">
+        <div className="col-12 justify-content-center align-items-center d-flex">
+          <div className="row row-eq-height">
+            <div className="col-lg-6 panel__content">
+              <table>
```
iscekic: Extract this `table` into a component and reuse it.

It should receive a config and render its rows based on what it received.

web/src/main/frontend/src/components/Category.js
```diff
@@ -0,0 +1,120 @@
...
+                  </tr>
+                </tbody>
+              </table>
+            </div>
+            <div className="col-lg-6 panel__content">
+              <div className="content__inner justify-content-center d-flex">
+                <div className="heading__row">
+                  <h2>Add/edit category</h2>
+                </div>
+                <div className="row input__block">
```
iscekic: Is this `input__block` needed?

web/src/main/frontend/src/components/Category.js
```diff
@@ -0,0 +1,120 @@
...
+                  <div className="col-12">
+                    <div className="row input__block_row">
```
iscekic: Extract this into a component and reuse it.

web/src/main/frontend/src/components/Category.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+
+class Category extends React.Component {
+  render() {
+    return (
+      <div className="row dashboard__row">
```
iscekic: Will the left part of the page **always** be the same for both `/category` and `/formula`?

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -0,0 +1,86 @@
...
+      border-bottom: 2px solid gray;
+      position: static;
+    }
+  }
+  tbody tr:nth-child(odd) {
+    background: #f2f2f2;
+  }
+  tr {
+    display: block;
+    margin-bottom: 0.625em;
```
iscekic: Use `px`.

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -0,0 +1,86 @@
...
+  }
+  border-collapse: collapse;
+  margin: 0;
+  padding: 0;
+  width: 100%;
+  table-layout: fixed;
+  color: $primary-dark-color;
+  thead {
+    border: none;
+    clip: rect(0 0 0 0);
```
iscekic: Is this needed?

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -0,0 +1,86 @@
...
+    height: 1px;
+    margin: -1px;
```
iscekic: Is this needed?

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -0,0 +1,86 @@
...
+  tbody tr:nth-child(odd) {
+    background: #f2f2f2;
+  }
+  tr {
+    display: block;
+    margin-bottom: 0.625em;
+    @media (min-width: 600px) {
+      margin-bottom: 0;
+      display: table-row;
+      padding: 0.35em;
```
iscekic: Use `px`.

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -0,0 +1,86 @@
...
+      @media (min-width: 600px) {
+        border-right: 2px solid gray;
+      }
+    }
+  }
+  th {
+    font-size: 16px;
+    font-weight: 500;
+    &:first-child {
+      width: 30%;
```
iscekic: I'm assuming this is due to design, right?

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -0,0 +1,86 @@
...
+    text-align: center;
+    &:first-child {
+      @media (min-width: 600px) {
+        border-right: 2px solid gray;
+      }
+    }
+  }
+  th {
+    font-size: 16px;
+    font-weight: 500;
```
iscekic: Use `bold` (or `bolder` or `normal`, whichever is the equivalent).

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -0,0 +1,86 @@
...
+    &:first-child {
+      width: 30%;
+    }
+  }
+  td {
+    display: block;
+    font-size: 0.8em;
```
iscekic: Use `px`.

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -0,0 +1,86 @@
...
+    text-align: center;
+    @media (min-width: 450px) {
+      text-align: right;
+    }
+    &:before {
+      content: attr(data-label);
```
iscekic: I strongly dislike this. Find a way to achieve the same without data attributes.

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -0,0 +1,86 @@
...
+  }
+  th {
+    font-size: 16px;
+    font-weight: 500;
+    &:first-child {
+      width: 30%;
+    }
+  }
+  td {
+    display: block;
```
iscekic: Why use a table and then change `td` display? Either don't touch it or create your own fake table with `divs`.

web/src/main/frontend/src/scss/components/_formula.scss
```diff
@@ -0,0 +1,31 @@
...
+    @media (min-width: 1199px) {
+      width: 40%;
+    }
+  }
+  .full__input {
+    width: 100%;
+  }
+  .number__input {
+    text-align: center;
+    padding: 2.5px 5px 2.5px 20px;
```
iscekic: `2.5px`? Are we really that precise? :smile:

web/src/main/frontend/src/components/Category.js
iscekic: If yes, extract it into a component and use it in `PanelContainer`.

web/src/main/frontend/src/scss/components/_category.scss
aberonja: I change it only on mobile cause it needs to be full width. After 600px it goes back to table-cell. Do you want me to find another way to make it full width?

iscekic: Please try and let me know what you came up with before committing.

web/src/main/frontend/src/scss/components/_category.scss
aberonja: Yes it is. Is it okay?

web/src/main/frontend/src/components/Category.js
aberonja: You're asking me why do I have input__block class? If that's the question it's because it adds padding, width and border on inputs

iscekic: I might've misread something... Carry on :smile:

web/src/main/frontend/src/scss/components/_category.scss
aberonja: I did some research and almost everywhere I looked they're using display block for responsive. I would leave it like this. Do you agree?

iscekic: Leave it like this then. :smile:

web/src/main/frontend/src/components/PanelContainerInner.js
```diff
@@ -0,0 +1,122 @@
...
+  render() {
+    return (
+      <div className="row dashboard__row">
+        <div className="col-12 justify-content-center align-items-center d-flex">
+          <div className="row row-eq-height">
+            <div className="col-lg-6 panel__content">
+              <Table heading={this.props.name} />
+            </div>
+            {this.props.name === 'Category' ? (
+              <div className="col-lg-6 panel__content">
```
aberonja: Would you like me to make two components instead of writing it all here?

web/src/main/frontend/src/components/PanelContainer.js
```diff
@@ -9,11 +11,16 @@ class PanelContainer extends React.Component {
...
         </div>
-        <div className="row dashboard__row">
-          <div className="col-12 dashboard__content">
-            <div>{this.props.name} content</div>
-          </div>
-        </div>
+        <Switch>
+          <Route
+            path={`/category`}
+            render={props => <PanelContainerInner name="Category" />}
```
iscekic: This is *kind of* ugly and also suggested against in the `react-router` docs (https://reacttraining.com/react-router/web/guides/basic-components/route-rendering-props). See my comment below.

web/src/main/frontend/src/components/PanelContainerInner.js
iscekic: You read my mind. :smile: It should be possible to leverage the router fully to get what we need, eg:

```js
// PanelContainer.jsx
<Switch>
   // no render fn, nice!
   render={PanelContainerInner}
</Switch>

// PanelContainerInner.jsx
// to render what we want we can:
// use props.match
// use <Switch>
// use the ternary operator
// explore the first two options and if they're not viable, use the third
```

web/src/main/frontend/src/components/SingleInput.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+class SingleInput extends React.Component {
+  render() {
+    return (
+      <div className={`row ${this.props.containerClass}`}>
```
iscekic: What if `this.props.containerClass` is `undefined`? :smile: Define `defaultProps`.

web/src/main/frontend/src/components/SingleInput.js
```diff
@@ -0,0 +1,27 @@
...
+            {this.props.name}
+          </label>
+        </div>
+        <div className="col-12 col-xl-9 text-left">
+          <input
+            id={this.props.name}
+            type="text"
+            placeholder={this.props.placeholder}
+            className={`input-inverted ${
+              this.props.inputClass !== undefined ? this.props.class : ''
```
iscekic: This is an ugly check. Define `defaultProps`.

web/src/main/frontend/src/components/Table.js
```diff
@@ -0,0 +1,56 @@
...
+    return (
+      <table>
+        <thead>
+          <tr>
+            <th scope="col">{this.props.heading}</th>
+            <th scope="col">Description</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
```
iscekic: These rows are repetitive. Make an array and iterate over it. :smile:

web/src/main/frontend/src/scss/layout/_panel.scss
```diff
@@ -101,8 +101,71 @@ $border: 1px solid white;
...
+            position: relative;
+            &:after {
+              content: '';
+              position: absolute;
+              bottom: -5px;
+              left: 50%;
+              right: 50%;
+              border-top: 1px solid $red-color;
+              width: 60%;
+              transform: translateX(-50%);
```
iscekic: Didn't we have a mixin for this?

web/src/main/frontend/src/components/PanelContainer.js
aberonja: I'm not sure that I understand what I need to do here. I separated category and formula into two diff components so I removed ternary operator. But this is suggested in `react-router` docs, I don't know what would you like me to do? :smile:

iscekic: Does this (https://github.com/execom-eu/OOP-Test-Generator/pull/67#discussion_r170594139) comment help?

web/src/main/frontend/src/components/PanelContainerInner.js
```diff
@@ -0,0 +1,40 @@
...
+  render() {
+    return (
+      <div className="row dashboard__row">
+        <div className="col-12 justify-content-center align-items-center d-flex">
+          <div className="row row-eq-height">
+            <div className="col-lg-6 panel__content">
+              <Table heading={this.props.name} />
+            </div>
+            <Switch>
+              <Route path="/category" component={Category} />
```
iscekic: So elegant. :100:

Issue #65, opened 2018-02-20
Pass answers to multiple choice questions when test is created


stanislavgrujic:
Update QuestionDto with a list of answers to be shown to the user, when a test is created, and pass them along the TestResponse.

PR #72, opened 2018-02-23
Bugfix/ndragic multiple choice answers

luinnarn:
Issue #65


web/src/main/java/eu/execom/testgenerator/web/dto/QuestionDto.java
```diff
@@ -1,23 +1,31 @@
...
-    private QuestionType type;
-    private String text;
-    private int points;
-    private byte[] image;
+  private Long id;
+  private QuestionType type;
+  private String text;
+  private int points;
+  private byte[] image;
+  private List<String> answers;
```
stanislavgrujic: Please try to think which of the fields from an `Answer` you might need in the frontend -> you need also the `id` of an answer, in order to be able to submit which answers user has selected.

Issue #68, opened 2018-02-21
Update Test class to handle max points and achieved points


stanislavgrujic:
Update the Test class with two new fields:
- maximumPoints
- achievedPoints

When a test is created, maximumPoints should be summed from included questions and populated in the database.
When a test is reviewed, achievedPoints should be populated. There is already a method `calculateTestScore` in the `TestService`, which should also update this field in the database.

PR #73, opened 2018-02-23
Feature/ndragic test maximum achieved points

luinnarn:
Issue #68


web/src/main/java/eu/execom/testgenerator/web/dto/TestResponse.java
```diff
@@ -16,6 +16,8 @@
   private Long applicantId;
   private Long reviewerId;
   private Long formulaId;
+  private int maximumPoints;
+  private int achievedPoints;
```
stanislavgrujic: Please add the xml schema updating the database explicitly.

core/src/main/resources/db/changelog/db.changelog-initializeDatabaseSchema.xml
```diff
@@ -37,6 +37,12 @@
       <column name="time_taken" type="timestamp">
         <constraints nullable="false"/>
       </column>
+      <column name="maximum_points" type="int">
+        <constraints nullable="true"/>
+      </column>
+      <column name="achieved_points" type="int">
+        <constraints nullable="true"/>
+      </column>
```
stanislavgrujic: At the moment, it is OK to leave these changes where you've put them, but you should avoid doing it. The reasoning -> if there is an application in production, it Flyway will have this script already executed (during the initial application startup) and it will never run again, therefore, it will never update the database.

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -126,11 +132,14 @@ public void remove(Long id) {
...
-               .mapToInt(i -> i)
-               .sum();
+    int achievedPoints = test.getTestQuestions()
+                             .stream()
+                             .map(testQuestion -> determineEarnedPoints(testQuestion, awardedPoints))
+                             .mapToInt(i -> i)
+                             .sum();
+    test.setAchievedPoints(achievedPoints);
+    testRepository.save(test);
+    return achievedPoints;
```
stanislavgrujic: Please add/update unit tests.

PR #81, opened 2018-02-28
Added routing in PanelContainer.js instead of just rendering component

aberonja:
In PanelContainer.js added switch instead of just rendering component on every /route
Removed PanelContainerInner.js
Added PanelContainerTwoCols.js for components in panel that have two cols
In StartScreen.js changed bootstrap classes for better responsive
In Panel.js changed bootstrap classes for better responsive
In _startLayout.scss changed height of background


web/src/main/frontend/src/components/PanelContainer.js
```diff
@@ -10,7 +11,20 @@ class PanelContainer extends React.Component {
             <h1>{this.props.name}</h1>
           </div>
         </div>
-        <PanelContainerInner name={this.props.name} />
+        <div className="row dashboard__row">
+          <div className="col-12 justify-content-center align-items-center d-flex">
+            <Switch>
+              <Route
+                path="/category"
+                render={props => <PanelContainerTwoCols name="Category" />}
```
iscekic: Remind me again, why can't we have just `PanelContainerTwoCols` here? :smile:

aberonja: Because there will be components that aren't going to be separated in two cols, some of them are going to be in one column. There are 3 components that are separated in two cols and the rest of them(which is also 3) are in a single col :smile:

Issue #88, opened 2018-03-02
Incorrect implementation of multiple choice answer validation


Varadski:
`public List<Answer> getSelectedAnswers() {
    return testQuestionAnswers.stream()
                              .filter(testQuestionAnswer -> testQuestionAnswer.isSelected())
                              .map(TestQuestionAnswer::getAnswer)
                              .collect(Collectors.toList());
  }`

The above will filter out the `TestQuestionAnswers` which were not selected, but it does not check if the selected answers are correct.

`public boolean isCorrectlyAnswered() {
    return getSelectedAnswers().stream().map(Answer::isCorrect).count() == question.getNumberOfCorrectAnswers();
  }`

This will count the number of selected answers that were provided by `getSelectedAnswers()` that were correct and if the numbers match it will return `true` and it will ignore all other answers which were provided (incorrect answers); it doesn't care how many answers were selected, or if incorrect answers were selected, as long as the number of correct answers selected is equal to the expected number of correct answers.

millevlada:
seems this is long ago merged

PR #89, opened 2018-03-03
Resolved issue #88.

Varadski:
Also changed `getSelectedAnswers()` filter to use a method reference instead of a lambda to make the code more concise.


core/src/main/java/eu/execom/testgenerator/core/model/TestQuestion.java
```diff
@@ -42,13 +42,15 @@
...
+                              .filter(TestQuestionAnswer::isSelected)
                               .map(TestQuestionAnswer::getAnswer)
                               .collect(Collectors.toList());
   }
 
   public boolean isCorrectlyAnswered() {
-    return getSelectedAnswers().stream().map(Answer::isCorrect).count() == question.getNumberOfCorrectAnswers();
+    int numOfSelected = getSelectedAnswers().size();
+    int numOfCorrect = (int) getSelectedAnswers().stream().map(Answer::isCorrect).count();
+    return  (numOfCorrect == question.getNumberOfCorrectAnswers() && numOfCorrect == numOfSelected);
```
stanislavgrujic: Please fix spacing and add a unit test for this.

PR #92, opened 2018-03-06
Updated Formula response/request classes.

MBarjaktarovic:


web/src/main/java/eu/execom/testgenerator/web/dto/GetAllFormulasResponse.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.List;
+
+@NoArgsConstructor
+@Data
+public class GetAllFormulasResponse {
+
+  private List<FormulaDto> formulaResponses;
```
stanislavgrujic: Please rename to `formulas`.

web/src/main/java/eu/execom/testgenerator/web/dto/TestDto.java
```diff
@@ -0,0 +1,40 @@
...
+import eu.execom.testgenerator.core.model.TestQuestion;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+@NoArgsConstructor
+@Data
+public class TestDto {
```
stanislavgrujic: Why is this class here?

web/src/main/java/eu/execom/testgenerator/web/dto/GetAllFormulasResponse.java
```diff
@@ -9,10 +9,10 @@
 @Data
 public class GetAllFormulasResponse {
 
-  private List<FormulaDto> formulaResponses;
+  private List<FormulaDto> formulasResponses;
```
stanislavgrujic: Just `formulas`. :)

PR #96, opened 2018-04-04
Moved Table to Separate components and Moved controls to PanelControls Component

vrackovic:
- Moved Table component render from PanelContainerTwoCells to Formula and Category as separate calls
- Outsourced panel controls to separate component
- Outsourced tableData to parent component


web/src/main/frontend/config-overrides.js
```diff
@@ -0,0 +1,9 @@
+const { injectBabelPlugin } = require("react-app-rewired");
+const rewireMobX = require("react-app-rewire-mobx");
+
+module.exports = function override(config, env) {
+  config = injectBabelPlugin("babel-plugin-styled-components", config);
+  config = rewireMobX(config, env);
+
+  return config;
+};
```
iscekic: Missing newline - was this file formatted with prettier?

web/src/main/frontend/src/base/Store.js
```diff
@@ -1,3 +1,40 @@
...
+    {title: 'Web Test', text: 'Test containing JS, HTML and CSS questions'},
+    {title: 'C# Test', text: 'Test containing C# and OOP questions'},
+    {title: 'Java Test', text: 'Test containing Java and OOP questions'},
+    {title: 'Android Test', text: 'Test containing...'}
+  ];
+
+  @computed
+  get category() {
+    return this.categories.map(c => {
+      c.id = Math.floor(Math.random() * 1000);
```
iscekic: Don't mutate the original array in `map`. Return a completely new item. :smile:

web/src/main/frontend/src/base/Store.js
```diff
@@ -1,3 +1,40 @@
...
+      return c;
+    });
+  }
+
+  @computed
+  get formula() {
+    return this.formulas.map(f => {
+      f.id = Math.floor(Math.random() * 1000);
```
iscekic: Same here.

web/src/main/frontend/src/components/ui/AddButton.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const button = props => {
+  return (
+    <div>
+      <div href={props.href} className={props.styling}>
```
iscekic: Rename `styling` to `className`. If this isn't possible because `className` is special, rename it to something else, eg. `cssClass`.

web/src/main/frontend/src/components/ui/Button.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const button = props => {
```
iscekic: Why do we have `Button` and `AddButton` as separate components? They are mostly the same.

web/src/main/frontend/src/components/ui/Table.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+class Table extends React.Component {
+  render() {
+    const tableRows = [];
```
iscekic: `tableRows` can be a computed class property.

web/src/main/frontend/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "babel-plugin-styled-components": "^1.5.1",
```
iscekic: Why was this added? I don't see any `styled-components`. :smile:

web/src/main/frontend/config-overrides.js
vrackovic: Don't think so. Newline will be added.

web/src/main/frontend/package.json
vrackovic: It will be removed :D

web/src/main/frontend/src/components/ui/Button.js
vrackovic: For now i solved it with combining both components, and render different things via ternary operator.

iscekic: There should be no ternary operator. Have the component render a button based on props, and handle everything else in the parent.

web/src/main/frontend/src/base/Store.js
```diff
@@ -1,3 +1,46 @@
...
+    {title: 'Web Test', text: 'Test containing JS, HTML and CSS questions'},
+    {title: 'C# Test', text: 'Test containing C# and OOP questions'},
+    {title: 'Java Test', text: 'Test containing Java and OOP questions'},
+    {title: 'Android Test', text: 'Test containing...'}
+  ];
+
+  @computed
+  get category() {
+    return this.categories.map(c => {
+      let computedCategory = {
```
iscekic: I'm assuming we're generating a random id to use as `key`? Do these objects not have a unique property which can be used?

If not, extract this callback into a separate function, eg.

```js
const addRandomId = obj => ({ ...obj, id: Math.floor(Math.random() * 1000) });

// elsewhere
@computed
get category() {
  return this.categories.map(addRandomId);
}
```

web/src/main/frontend/src/components/ui/Table.js
```diff
@@ -0,0 +1,32 @@
...
+
+    return (
+      <table>
+        <thead>
+          <tr>
+            <th scope="col">{this.props.heading}</th>
+            <th scope="col">Description</th>
+          </tr>
+        </thead>
+        {this.props.tableData ? tableData : null}
```
iscekic: No need for this, just `{tableData}` should be enough.

web/src/main/frontend/src/base/Store.js
vrackovic: There are two objects with same `title` property, and it was used for assigning `key`-s.

PR #106, opened 2018-05-08
Jobradovic done test store

jolpazy:


web/src/main/frontend/src/components/MultipleChoiceQuestion.js
```diff
@@ -5,31 +5,28 @@ import Checkbox from './ui/Checkbox';
 @observer
 class MultipleChoiceQuestion extends Component {
   render() {
+    let options = this.props.options.map(option => (
+      <div className="row row-eq-height answer" key={option.id}>
+        <div className="col-2">
+          <Checkbox
+            checked={
+              this.props.solution && this.props.solution === option.id
```
iscekic: `this.props.solution === option.id` should be enough.

web/src/main/frontend/src/components/ReviewTest.js
```diff
@@ -1,10 +1,50 @@
...
 
+@inject('completedTest')
+@observer
 class ReviewTest extends Component {
+  questionsToComponents = (question, index) => {
+    switch (question.type) {
+      case 'CodeChallenge':
+        return (
+          <CodeChallengeQuestion
+            index={index + 1}
```
iscekic: Why is this here?

web/src/main/frontend/src/components/MultipleChoiceQuestion.js
jolpazy: still need to return null and not false in case there is no solution passed, since false on checked property makes it read only, and returning null means there's no checked property at all

web/src/main/frontend/src/components/ReviewTest.js
jolpazy: So the first question would be 1. not 0. in the UI

web/src/main/frontend/src/components/MultipleChoiceQuestion.js
jolpazy: Idea is that if we pass solution to MultipleChoice question, it makes it's checkboxes readonly with checked and disabled properties, if not, it is normal checkbox

web/src/main/frontend/src/components/MultipleChoiceQuestion.js
```diff
@@ -5,31 +5,24 @@ import Checkbox from './ui/Checkbox';
 @observer
 class MultipleChoiceQuestion extends Component {
   render() {
+    let options = this.props.options.map(option => (
+      <div className="row row-eq-height answer" key={option.id}>
+        <div className="col-2">
+          <Checkbox
+            checked={this.props.solution === option.id ? true : null}
```
iscekic: `this.props.solution === option.id || null` and we can merge. :smile:

PR #107, opened 2018-05-08
Vrackovic creating take test page

vrackovic:
- Created TakeTest component
- Implemented Checkbox component
- Implemented onScroll event that adds top and bottom shadows in test components


web/src/main/frontend/src/components/TakeTest.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+
+import MultipleChoiceQuestion from './MultipleChoiceQuestion';
+import TextAreaQuestion from './TextAreaQuestion';
+import CodeChallengeQuestion from './CodeChallengeQuestion';
+
+class TakeTest extends React.Component {
+  state = {
+    shadows: undefined
```
iscekic: No need to explicitly define it as `undefined`.

iscekic: Initial value should be `''` I guess.

web/src/main/frontend/src/components/TakeTest.js
```diff
@@ -0,0 +1,79 @@
...
+  };
+
+  componentDidMount() {
+    this.handleDisplayShadow(this.testArea);
+  }
+
+  handleDisplayShadow(targetElement) {
+    const shouldShowBottomShadow =
```
iscekic: No need for should in the variable name. `showBottomShadow` is enough. :smile:

web/src/main/frontend/src/components/TakeTest.js
```diff
@@ -0,0 +1,79 @@
...
+      targetElement.scrollHeight -
+        targetElement.scrollTop -
+        targetElement.offsetHeight >
+      10;
+    const shouldShowTopShadow = targetElement.scrollTop > 10;
+    const shouldShowBothShadows = shouldShowTopShadow && shouldShowBottomShadow;
```
iscekic: There's no need to handle the both case. Everything below this line can be as simple as:

```js
this.setState({
  shadows: `${showBottomShadow ? '...' : ''} ${showTopShadow ? '...' : ''}`.trim()
});
```

web/src/main/frontend/src/components/TakeTest.js
```diff
@@ -0,0 +1,79 @@
...
+    } else if (shouldShowBottomShadow) {
+      this.setState({
+        shadows: 'test__bottom_shadow'
+      });
+    }
+  }
+
+  handleScroll = event => {
+    event.stopPropagation();
+    const targetElement = event.target;
```
iscekic: Don't create a variable here, scroll events fire very often.

web/src/main/frontend/src/scss/base/_base.scss
```diff
@@ -1,6 +1,7 @@
 body {
   -webkit-font-smoothing: antialiased;
   font-family: $base_font;
+  user-select: none;
```
iscekic: Double check that this is intended.

web/src/main/frontend/src/scss/components/_reviewTest.scss
```diff
@@ -3,6 +3,7 @@
   margin-left: 10px;
   border: 1px solid black;
   margin-bottom: 5px;
+  outline: none;
```
iscekic: Lots of `outline: none`, double check if they're all needed and if it may be done in a more DRY way.

web/src/main/frontend/src/scss/components/_takeTest.scss
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+.test__top_shadow {
+  box-shadow: inset 0px 10px 9px -7px rgba(0, 0, 0, 0.4);
+}
+.test__bottom_shadow {
+  box-shadow: inset 0px -10px 9px -7px rgba(0, 0, 0, 0.4);
+}
+.test__both_shadows {
```
iscekic: This can be removed.

web/src/main/frontend/src/scss/components/_takeTest.scss
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.take-test__background {
+  background: white;
+  margin: 15% !important;
+  padding: 20px 40px;
+  box-shadow: 0px 0px 15px 5px grey;
+
+  @media (max-width: 1200px) {
+    margin: 10% !important;
```
iscekic: Please try to avoid using `!important`. It occurs in too many places in this file for my taste.

web/src/main/frontend/src/scss/base/_base.scss
vrackovic: It is. Idea is that all content will be read-only because copy-pasting could help applicants to cheat. :)

iscekic: I have a better idea. You should be able to intercept the copy event on the body and make the copied text always `No cheating! :)`.

Some people read with their mouse, and disabling text selection makes it awkward for them.

web/src/main/frontend/src/scss/components/_takeTest.scss
vrackovic: Temporary solution that somehow finds its way into commit :smile:

web/src/main/frontend/src/components/TakeTest.js
```diff
@@ -0,0 +1,81 @@
...
+import CodeChallengeQuestion from './CodeChallengeQuestion';
+
+class TakeTest extends React.Component {
+  state = {
+    shadows: ''
+  };
+
+  componentDidMount() {
+    this.handleDisplayShadow(this.testArea);
+    document.addEventListener('copy', this.handleCopyEvent);
```
iscekic: This shouldn't be in this component, it should be as close to the entry component as possible. Probably in the entry component itself.

web/src/main/frontend/src/index.js
```diff
@@ -11,17 +11,23 @@ import FormulaStore from './base/stores/FormulaStore';
 import QuestionStore from './base/stores/QuestionStore';
 
 class Root extends React.Component {
+  handleCopyEvent(e) {
```
iscekic: Maybe it's better for this to be in `App`. No extra `div` created that way. :smile:

iscekic: @vrackovic Actually, leave it as is. Resolve confilcts and we can merge. :smile:

PR #112, opened 2018-05-11
Jobradovic point rework

jolpazy:


web/src/main/frontend/src/components/utility/getComponentByQuestionType.js
```diff
@@ -49,6 +49,9 @@ const getComponentByQuestionType = (props, index) => {
             question={props.question}
             options={props.options}
             solution={props.solutionID}
+            isCorrect={
```
vrackovic: Will `Array.prototype.find()` break if called on undefined?

web/src/main/frontend/src/components/Points.js
```diff
@@ -11,32 +11,46 @@ class Points extends React.Component {
 
   createInputField(fixed, correct, valuated) {
     const className = fixed && !correct ? 'no-spinner wrong' : 'no-spinner';
-    const value = fixed ? this.props.points : null;
+    const value = fixed
```
vrackovic: There must be a better solution than double ternary :smile:

web/src/main/frontend/src/components/Points.js
```diff
@@ -11,32 +11,46 @@ class Points extends React.Component {
...
+        onFocus={onFocus}
       />
     );
   }
 
   turnOffScrollWheel = () => e => e.preventDefault();
 
+  onFocus = e => {
+    if (e.target.value && e.target.value.includes('/')) {
+      e.target.value = parseInt(e.target.value.split('/')[0], 10);
```
iscekic: We shouldn't modify `value` directly.

web/src/main/frontend/src/components/Points.js
```diff
@@ -11,32 +11,46 @@ class Points extends React.Component {
...
+    }
+  };
+
   onChange = event => {
     let value = parseInt(event.target.value, 10);
 
     if (isNaN(value) || value <= 0) {
       value = 0;
-      event.target.value = null;
+      event.target.value = '';
```
iscekic: Same here.

web/src/main/frontend/src/components/Points.js
```diff
@@ -47,6 +61,12 @@ class Points extends React.Component {
     this.props.completedTest.addScore(value, this.props.id);
   };
 
+  onBlur = event => {
+    if (event.target.value) {
+      event.target.value = event.target.value + `/${this.props.points}`;
```
iscekic: Here as well.

web/src/main/frontend/src/components/Points.js
jolpazy: It looks prettier, and more readable when properly formatted, butt I'll extract it into a method anyway

web/src/main/frontend/src/components/ui/Checkbox.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 
 const checkbox = props => {
+  const answerTextClassName =
+    props.checked && !props.isCorrect ? 'label_text wrong' : 'label_text';
```
iscekic: Use a mutable variable, set it to `label_text` and then conditionally add `wrong`.

web/src/main/frontend/src/components/utility/getComponentByQuestionType.js
iscekic: It will. Will we ever get `undefined` here is the better question. 😄

vrackovic: We probably will because applicants shouldn't get solutions

web/src/main/frontend/src/components/ui/Checkbox.js
iscekic: Doesn't even need to be mutable:

```js
const answerTextClassName = `label_text ${props.checked && !props.isCorrect ? 'wrong' : ''}`;
```

web/src/main/frontend/src/components/utility/getComponentByQuestionType.js
jolpazy: My idea was that multiple choice question component gets the isCorrect prop only when used inside reviewTest component

web/src/main/frontend/src/components/Points.js
jolpazy: How should i modify the value of input? One of the use cases is that i want to set it to 0 if user tries tou enter less than 0 or I want to fall back to placeholder if input is cleared

iscekic: Make it a controlled input imo - https://reactjs.org/docs/forms.html#controlled-components

web/src/main/frontend/src/components/utility/getComponentByQuestionType.js
jolpazy: Added check before looking for isCorrect

web/src/main/frontend/src/components/Points.js
```diff
@@ -3,57 +3,103 @@ import {inject} from 'mobx-react';
...
 
-  createInputField(fixed, correct, valuated) {
-    const className = fixed && !correct ? 'no-spinner wrong' : 'no-spinner';
-    const value = fixed ? this.props.points : null;
-    const readOnly = fixed || null;
-    const onWheel = valuated && this.turnOffScrollWheel;
-    const onChange = valuated && this.onChange;
-    const placeholder = valuated ? this.props.points : null;
+  getValueForInput = (fixed, correct) => {
+    if (fixed) {
```
iscekic: I think this will look cleaner if we extract `inputValue`:

```js
let inputValue = '';
if (fixed) {
  inputValue = correct ? ... : ... ;
}

this.setState({ inputValue });
```

web/src/main/frontend/src/components/Points.js
```diff
@@ -3,57 +3,103 @@ import {inject} from 'mobx-react';
...
+    inputProps.onWheel = valuated && this.turnOffScrollWheel;
+    inputProps.onChange = valuated && this.onChange;
+    inputProps.onBlur = valuated && this.onBlur;
+    inputProps.onFocus = valuated && this.onFocus;
+    inputProps.onKeyDown = valuated && this.onKeyDown;
+    inputProps.placeholder = `0/${this.props.points}`;
+    return inputProps;
   }
 
   turnOffScrollWheel = () => e => e.preventDefault();
```
iscekic: I know that this isn't part of this PR, but why is this `() => e ...` instead of `e => ...`?

web/src/main/frontend/src/components/Points.js
```diff
@@ -3,57 +3,103 @@ import {inject} from 'mobx-react';
 
 @inject('completedTest')
 class Points extends React.Component {
+  state = {
+    inputValue: ''
+  };
+
+  reference = React.createRef();
```
iscekic: Rename to `inputRef`.

web/src/main/frontend/src/components/ui/Checkbox.js
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 
 const checkbox = props => {
+  const answerTextClassName = `label_text ${
+    props.checked && !props.isCorrect ? 'wrong' : ''
+  }`;
```
iscekic: `\n`

web/src/main/frontend/src/components/Points.js
```diff
@@ -3,57 +3,103 @@ import {inject} from 'mobx-react';
...
+
   componentDidMount() {
     if (this.props.isCorrect) {
       this.props.completedTest.addScore(this.props.points, this.props.id);
     }
+    this.getValueForInput(this.props.fixedScore, this.props.isCorrect);
```
iscekic: `\n` above

web/src/main/frontend/src/components/Points.js
```diff
@@ -3,57 +3,103 @@ import {inject} from 'mobx-react';
...
+      fixed && !correct ? 'wrong' : ''
+    }`;
+    inputProps.readOnly = fixed || null;
+    inputProps.onWheel = valuated && this.turnOffScrollWheel;
+    inputProps.onChange = valuated && this.onChange;
+    inputProps.onBlur = valuated && this.onBlur;
+    inputProps.onFocus = valuated && this.onFocus;
+    inputProps.onKeyDown = valuated && this.onKeyDown;
+    inputProps.placeholder = `0/${this.props.points}`;
+    return inputProps;
```
iscekic: `\n` above

web/src/main/frontend/src/components/Points.js
jolpazy: not needed anymore, now that i think about it

PR #113, opened 2018-05-11
Vrackovic added placeholders to select elements

vrackovic:
- Added placeholder for HTML Select elemenets
- Added local state to change selected option color


web/src/main/frontend/src/components/ui/Select.js
```diff
@@ -1,12 +1,27 @@
...
 class Select extends React.Component {
+  state = {
+    color: 'grey'
+  };
+
+  handleColor = () => {
+    this.setState({
+      color: this.props.color
+    });
+  };
```
iscekic: `\n`

web/src/main/frontend/src/components/ui/Select.js
```diff
@@ -1,12 +1,27 @@
...
   render() {
-    const options = this.props.options.map(option => (
+    const optionsWithPlaceholder = addOptionsPlaceholder(
+      this.props.options,
+      this.props.placeholder
+    );
```
iscekic: `\n`

web/src/main/frontend/src/components/utility/addOptionsPlaceholder.js
```diff
@@ -0,0 +1,4 @@
+const addOptionsPlaceholder = (optionsArray, placeholder) =>
```
iscekic: Will this ever be used outside of `Select`? If not, make it a method of that component.

vrackovic: It will not. Will be made into method

web/src/main/frontend/src/components/ui/Select.js
```diff
@@ -1,12 +1,27 @@
 import React from 'react';
+import addOptionsPlaceholder from '../utility/addOptionsPlaceholder';
 
 class Select extends React.Component {
+  state = {
+    color: 'grey'
```
jolpazy: Don't use color names, as it may get interpreted differently by different browsers (:

PR #115, opened 2018-05-15
Jobradovic header counter

jolpazy:


web/src/main/frontend/src/components/Test.js
```diff
@@ -7,8 +7,24 @@ class Test extends React.Component {
 
   componentDidMount() {
     this.handleDisplayShadow(this.testArea);
+    this.elements = document.getElementsByClassName('question-div');
```
iscekic: Use a `ref`.

jolpazy: Elements that I'm trying to grab here are not a part of this component, question-div's are generated form another function outside. Therefore document.get as a solution. Any ideas?

iscekic: Hmmm, I guess we can't use refs since we have multiple elements we need to loop through. So this approach is fine. Very ugly though.

Do we really need this functionality? Was it explicitly requested? 😄

jolpazy: Was aware of the ugliness as I made it (: 
Question step counter was implicitly proposed through mock-up images

iscekic:
Ok, you can merge this, but if we ever have issues with it we will remove it. 😄

PR #134, opened 2018-05-30
Jobradovic questions api

jolpazy:


web/src/main/frontend/src/base/stores/QuestionStore.js
```diff
@@ -1,11 +1,22 @@
...
 import * as util from '../util.js';
+import questionApi from '../rest/questionApi';
 
 class QuestionStore {
   transformTable = util.tableTransformer;
 
   @observable questions = [];
 
+  // Temp solution for getting only the values we want to display
+  questionCutter = q => ({text: q.text, points: q.points, id: q.id});
```
iscekic: ✂️ Why is this a temp solution? ✂️

jolpazy: Or we can just leave it and edit later if we want to show more values.

iscekic: Remove the comment then. Also maybe think of changing the name. 💃

PR #135, opened 2018-05-30
Vrackovic/submitting category form

vrackovic:
- Submiting category form functionality added with Add/Edit mods
- Table changed to support row selection
- PanelControls now work as intended
- `description` field supported. Submitting it would have no result, so it should work when `description` is added to DB


web/src/main/frontend/src/base/stores/CategoryStore.js
```diff
@@ -5,13 +5,37 @@ import categoryApi from '../rest/categoryApi';
...
   fetchCategories() {
     return categoryApi
       .getAllCategories()
       .then(categories => (this.categories = categories));
   }
 
+  @action.bound
+  submitCategory() {
+    if (this.categoryForm.id) {
+      return categoryApi.updateCategory(this.categoryForm).then(category => {
```
jolpazy: No need to pass category to this.fetchCategories();

web/src/main/frontend/src/base/rest/categoryApi.js
```diff
@@ -1,9 +1,17 @@
-import {get, withLoader} from './requests';
+import {get, post, put, withLoader} from './requests';
 
 class CategoryApi {
   getAllCategories() {
     return withLoader(get('/categories')).then(res => res.categoryResponses);
   }
+
+  createNewCategory(newCategory) {
+    return post('/categories').send(newCategory);
```
iscekic: Should create/update use `withLoader`?

web/src/main/frontend/src/components/Category.js
```diff
@@ -12,6 +12,22 @@ class Category extends React.Component {
     this.props.category.fetchCategories();
   }
 
+  handleChange = event => {
+    this.props.category.setCategoryForm({
```
iscekic: Please investigate https://github.com/foxhound87/mobx-react-form - would it be overkill? @jobradovicexecom @sugrinov

web/src/main/frontend/src/components/ui/TableBody.js
```diff
@@ -1,11 +1,14 @@
 import React from 'react';
-import TrashIcon from '../../img/trash.png';
-import EditIcon from '../../img/edit.png';
 import capitalize from '../utility/capitalize';
 
-export function TableBody({tableHeadings, tableRows, isAdmin}) {
+export function TableBody({tableHeadings, tableRows, isAdmin, rowClick}) {
```
iscekic: Instead of a ternary, would it be possible to set a default value for `rowClick`?

```
... , rowClick = () => {}}) {
```

web/src/main/frontend/src/scss/components/_category.scss
```diff
@@ -15,6 +15,14 @@ table {
       display: table-header-group;
     }
   }
+  tbody {
+    tr:hover {
+      color: #fff;
+      cursor: pointer;
+      background-color: #3c3c46 !important;
```
iscekic: `!important`? 😄

web/src/main/frontend/src/components/ui/TableBody.js
vrackovic: Yes, it is possible and will be done :smile:

web/src/main/frontend/src/scss/components/_category.scss
vrackovic: Not really :smile:

web/src/main/frontend/src/components/Category.js
sugrinov: I think it it's not an overkill since it at least abstracts away `form states` which may prove useful.

iscekic: I agree, I think it will abstract away (arguably) ugly form handling code. @jobradovicexecom What do you think? 😄

jolpazy: It does seem like a good (and complex) library. We can use it for the sake of learning it. Even though i like manual javascript validation (:

iscekic: Ok, lets use it then. 😄

jolpazy: Can we finish this PR up, and then implement the library in separate smaller ones?

PR #136, opened 2018-05-30
Jobradovic dashboard fetch

jolpazy:
Data for dashboard is fetched


web/src/main/frontend/src/base/rest/dashboardApi.js
```diff
@@ -0,0 +1,9 @@
+import {get, withLoader} from './requests';
+
+class DashboardApi {
+  getDashboardData() {
+    return withLoader(get('/dashboard')).then(res => res.dashboardTestDtos);
```
iscekic: `dashboardTestDtos` should probably be named differently on the BE. @sugrinov

PR #138, opened 2018-05-31
Mobx react form implementation for category

vrackovic:
- Category form submission and edit moved from store controlled to mobx-react-form
- Formula Form submission and edit done by @jolpazy


package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
```
iscekic: This isn't the right `package.json`. ⚠️

web/src/main/frontend/src/components/Category.js
```diff
@@ -4,35 +4,58 @@ import SingleInput from './ui/SingleInput';
...
+    console.log(form.touched);
+  }
+};
+
+const options = {
+  showErrorsOnReset: false,
+  validateOnBlur: false
+};
+
+const form = new MobxReactForm({fields}, {options, plugins, hooks});
```
iscekic: Move the form to a new file.

yarn.lock
```diff
@@ -0,0 +1,17 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
iscekic: This isn't the right `yarn.lock`. ⚠️

web/src/main/frontend/src/components/Category.js
```diff
@@ -64,38 +83,44 @@ class Category extends React.Component {
...
                 </div>
                 <div className="col-12">
                   <SingleInput
+                    formBind={form.$('description').bind()}
                     name="Description"
                     placeholder="Category description"
                     containerClass="input__block_row"
-                    inputValue={this.props.category.categoryForm.description}
-                    change={this.handleChange}
+                    validationError={form.$('description').error}
```
iscekic: These 2 props are unused at the moment?

web/src/main/frontend/src/components/forms/CategoryForm.js
```diff
@@ -0,0 +1,40 @@
...
+};
+
+const hooks = {
+  onSuccess(form) {
+    CategoryStore.submitCategory(form.values());
+    form.reset();
+  },
+
+  onError(form) {
+    form.invalidate('This is a generic error message!');
```
iscekic: I'm assuming this won't be changed as part of this PR. Leave a `// TODO`

web/src/main/frontend/src/components/forms/FormulaForm.js
```diff
@@ -0,0 +1,37 @@
...
+  onSuccess(form) {
+    FormulaStore.submitFormula(form.values()).then(() => form.clear());
+  },
+
+  onError(form) {
+    form.invalidate();
+  },
+
+  onSubmit() {
+    console.log(4);
```
iscekic: Same here.

web/src/main/frontend/src/components/utility/capitalize.js
```diff
@@ -1,4 +1,6 @@
 const capitalize = word => {
+  // Temporary solution for not breaking the app (:
+  if (!word) return;
```
iscekic: Return `''` instead of `undefined`.

web/src/main/frontend/src/components/Formula.js
```diff
@@ -30,23 +31,29 @@ class Formula extends React.Component {
...
               tableRows={this.props.formula.formulaTable.tableRows}
               headerRows={this.props.formula.formulaTable.headerRows}
+              rowClick={this.handleRowClick}
             />
             <div className="add__button_container">
               <Button cssClass="icon__block" title="+" />
             </div>
           </div>
-          <div className="col-lg-6 panel__content">
+          <form className="col-lg-6 panel__content" id="formulaForm">
```
iscekic: Is `id` needed here?

vrackovic: Nope, it was needed when we were submitting form without mobx-react-form

vrackovic: As it turns out, it is needed for being able to submit form with `Enter` key press.

web/src/main/frontend/src/components/utility/capitalize.js
vrackovic: Since form validation is implemented, this check is not needed, but will be left to return empty string just in case someone submits empty form in process of giving life to other forms. :laughing:

web/src/main/frontend/src/components/Formula.js
iscekic: Why is it needed for enter? It shouldn't be.

vrackovic: It is because submit button is outside of `<form>` tags. Should i wrap whole component with `<form>` or keep this version with `formId`?

iscekic: Wrap it imo

PR #141, opened 2018-06-01
Feature/#140 sugrinov add java configration frontend

sugrinov:
* Added java configuration for frontend
* Added api to controller request mappings


web/src/main/frontend/package.json
```diff
@@ -34,7 +34,7 @@
     "start-js": "react-app-rewired start",
     "start": "npm-run-all -p watch-css start-js",
     "build-js": "react-app-rewired build",
-    "build": "npm-run-all build-css build-js",
+    "build": "npm-run-all build-css build-js && cp build/* ../resources/static -rf",
```
iscekic: Keep in mind that `cp` isn't available on Windows - please solve it in a cross-platform way. 😄

web/src/main/resources/static/index.html
```diff
@@ -1,11 +1 @@
...
-    <meta charset="UTF-8">
-    <title>Index</title>
-</head>
-<body>
-    <h3>This is index, and everybody can see this.</h3>
-    <a href="register" data>login with google</a>
-</body>
-</html>
\ No newline at end of file
+<!DOCTYPE html><html lang="en"><head><meta charset="utf-8"><meta name="viewport" content="width=device-width,initial-scale=1,shrink-to-fit=no"><meta name="theme-color" content="#000000"><link rel="manifest" href="/manifest.json"><link rel="shortcut icon" href="/favicon.png"><link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/css/bootstrap.min.css" integrity="sha384-Gn5384xqQ1aoWXA+058RXPxPg6fy4IWvTNh0E263XmFcJlSAwiGgFAW/dAiS6JXm" crossorigin="anonymous"><title>Execom test</title><link href="/static/css/main.6f1d489f.css" rel="stylesheet"></head><body><noscript>You need to enable JavaScript to run this app.</noscript><div id="root"></div><script type="text/javascript" src="/static/js/main.3a9441e5.js"></script></body></html>
```
iscekic: Is this file needed?

web/src/main/frontend/package.json
sugrinov: You're right. I will update it

web/src/main/resources/static/index.html
sugrinov: Not really. I can put it in .gitignore, it will be created in build

iscekic: Please do. 😄 Also delete this file. 💃

core/src/main/resources/application.properties
```diff
@@ -29,6 +29,4 @@ security.oauth2.client.clientAuthenticationScheme=form
 security.oauth2.client.scope=profile email
 
 security.oauth2.resource.userInfoUri=https://www.googleapis.com/userinfo/v2/me
-security.oauth2.resource.preferTokenInfo=false 
-
-server.contextPath=/api
+security.oauth2.resource.preferTokenInfo=false
```
iscekic: nit: add `\n`

web/pom.xml
```diff
@@ -47,4 +48,50 @@
...
+            </goals>
+            <phase>prepare-package</phase>
+            <configuration>
+              <workingDirectory>${frontend.build.directory</workingDirectory>
+              <nodeVersion>v8.11.2</nodeVersion>
+            </configuration>
+          </execution>
+
+          <execution>
+            <id>npm install</id>
```
iscekic: Can we use `yarn`?

web/src/main/frontend/package.json
```diff
@@ -34,8 +35,9 @@
     "start-js": "react-app-rewired start",
     "start": "npm-run-all -p watch-css start-js",
     "build-js": "react-app-rewired build",
-    "build": "npm-run-all build-css build-js",
+    "build": "npm-run-all build-css build-js copy-build",
     "test": "react-app-rewired test --env=jsdom",
+    "copy-build": "./node_modules/.bin/n-copy --source build --destination ../resources/static \"**\" ",
```
iscekic: No need to prefix with `./node_modules/.bin`

web/pom.xml
sugrinov: At the moment no because of how yarn is installed (yarn/yarn-v{version}/) and how this plugin handles yarn (tries to find it in yarn/dist/bin/yarn) but it could be possible in version 1.7 there is a [PR](https://github.com/eirslett/frontend-maven-plugin/issues/647) about this but I don't know if and when it will be added.

core/src/main/resources/application.properties
sugrinov: (Y)

web/pom.xml
iscekic: Mili Bože čuda velikoga † 😆

sugrinov: :latin_cross:

PR #144, opened 2018-06-06
Revert changes done to authentication

sugrinov:
* Remove /api from controllers and add it back to application.properties
* Fix failing tests


sugrinov:
Previous changes break the flow for frontend developers when they develop on different ports so they are reverted.
I will implement auth on different branch from develop. @stanislavgrujic

web/src/main/java/eu/execom/testgenerator/web/security/RoleAuthenticationFilter.java
```diff
@@ -45,13 +47,19 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
-    String email = details.get("email");
-    User user = userService.findByEmail(email);
-    if (user == null || !user.getRole().hasAdministrativeRights()) {
-      SecurityContextHolder.clearContext();
-      ((HttpServletResponse) response).setStatus(HttpStatus.UNAUTHORIZED.value());
-      ((HttpServletResponse) response).sendRedirect("/");
-    } else {
+    // TODO sugrinov: 6.6.18. Added after discussion with vmijic. Look into moving this logic elsewhere so we don't have
+    // multiple queries for user on every request.
+    if (null == initialized.get()) {
```
stanislavgrujic: Please do not use inverted null checks...

PR #148, opened 2018-06-13
Questions form delete

vrackovic:
- Added delete button, store action and API call


web/src/main/frontend/src/components/ui/DeleteButton.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import Button from './Button';
+
+const deleteButton = props => {
+  return (
+    <div className="col-12 delete__button_container">
+      <div className="d-flex justify-content-end">
+        {props.id > 0 && (
```
iscekic: `props.id && ...` is enough.

iscekic: Are we only hiding the button, or do we want to hide the whole thing (container and all)?

jolpazy: In current implementation of showing/hiding delete button, id is set to 0 by clear function of mobx-react-form.
So props.id && ... returns 0

vrackovic: Yep, `props.id` is enough.
We could hide whole thing. When button is hidden, container is rendered as `div` with 100% width and no height.

iscekic: `!!props.id`? 😄

iscekic: I guess I'm nitpicking though, just leave this as is.

iscekic: @vrackovic Hide the entire container and we merge. 💯

vrackovic: @jolpazy Aren't we using `onReset`, not `onClear`?

jolpazy: Whatever we are using, it leaves an id = 0

PR #150, opened 2018-06-18
Formula component dynamic category selects

vrackovic:
- Category fields in formula components can be added or removed dynamically
- Implemented react-select
- Slightly changed styles


web/src/main/frontend/src/components/FormulaItem.js
```diff
@@ -0,0 +1,54 @@
...
+    const {id, changeCategory} = this.props;
+    this.setState({
+      selectedCategory: selection.value
+    });
+    changeCategory({...selection, id});
+  };
+
+  render() {
+    return (
+      <div className="formula__category_wrapper" key={this.props.id}>
```
iscekic: Why do we need `key` here?

web/src/main/frontend/src/components/FormulaItem.js
```diff
@@ -0,0 +1,54 @@
...
+@inject('formula')
+@observer
+class FormulaItem extends React.Component {
+  state = {
+    selectedCategory: ''
+  };
+
+  handleCategoryChange = selection => {
+    const {id, changeCategory} = this.props;
+    this.setState({
```
iscekic: Can we remove local state?

web/src/main/frontend/src/components/FormulaItem.js
vrackovic: We don't :)

web/src/main/frontend/src/components/FormulaItem.js
vrackovic: We can try :)

web/src/main/frontend/src/base/stores/FormulaStore.js
```diff
@@ -1,12 +1,37 @@
...
+      this.formCategories = this.formCategories.filter(
+        category => category.id !== categoryId
+      );
+    }
+  }
+
+  @action.bound
+  setFormCategories(category) {
+    this.formCategories = this.formCategories.map(
+      form => (form.id === category.id ? category : form)
```
iscekic: Rename `form` to `existingCategory`.

PR #151, opened 2018-06-20
Guest login form fix

vrackovic:
- Fixed react-select to consist with project styles
- Form submission functionality implemented
- User authentication moved to `Panel` component so anonymous users can access applicant form
- Added TestGeneratorDiagram to gitignore


web/src/main/frontend/src/pages/GuestLogin.js
```diff
@@ -10,45 +11,84 @@ class GuestLogin extends React.Component {
     this.props.category.fetchCategories();
   }
 
+  changeHandler = type => {
+    if (!type) {
```
iscekic: Re-check if this is needed.

web/src/main/frontend/src/pages/Panel.js
```diff
@@ -20,6 +27,19 @@ class Panel extends React.Component {
     {name: 'Login page', url: '/login', image: {name: LoginIcon}}
   ];
 
+  componentDidMount() {
+    this.props.auth.login().then(res => {
+      this.setState({
+        redirect:
+          res.status !== 200 ? (
```
iscekic: Invert this check: `res.status === 200 ? ...`

web/src/main/frontend/src/pages/Panel.js
```diff
@@ -41,12 +61,12 @@ class Panel extends React.Component {
               </nav>
             </div>
             <Switch>
-              <Redirect exact from="/register" to="/dashboard" />
+              {this.state.redirect}
```
iscekic: Only save `status` in the state and then render the redirect based on that.

web/src/main/frontend/src/scss/components/_guestLogin.scss
```diff
@@ -0,0 +1,104 @@
+.guest__form_fields {
```
iscekic: Please create an issue to clean up css in general.

web/src/main/frontend/src/pages/GuestLogin.js
vrackovic: As it turns out, this check is needed. Even when `react-select` field is specified as `clearable=false`, and `X` doesn't exist, it is possible to clear it with pressing `Backspace`

web/src/main/frontend/src/pages/Panel.js
vrackovic: We found a better implementation of `Oauth`, so this part will probably be removed soon

vrackovic:
Closing pull request until login is properly implemented

PR #164, opened 2018-07-02
Feature/sugrinov add description to formula

sugrinov:


web/src/test/java/eu/execom/testgenerator/web/FormulaMother.java
```diff
@@ -19,11 +21,26 @@ public static Formula formula(Long id, String name) {
...
+
   public static String formulaRequestJson(String name) {
     return "{\"name\" : \"" + name + "\"}";
   }
 
   public static String formulaRequestJson(String name, Long id) {
     return "{\"name\" : \"" + name + "\", \"id\":\"" + id + "\"}";
   }
+
+  public static String formulaRequestJson(String name, Long id, String description) throws JsonProcessingException {
```
stanislavgrujic: It is a good practice to wrap checked exceptions inside a method and throw a RuntimeException, so that you don't clutter the code outside.

PR #167, opened 2018-07-02
Formula description field frontend addition

vrackovic:
- Added description field to formula form
- Tested POST, PUT and DELETE methods, all working as expected


web/src/main/frontend/src/components/forms/FormulaForm.js
```diff
@@ -18,6 +22,7 @@ const hooks = {
   },
 
   onError(form) {
+    console.log(form.values());
```
iscekic: ⚗️

PR #168, opened 2018-07-03
Feature/#165 sugrinov wrap checked exceptions inside method

sugrinov:


web/src/main/java/eu/execom/testgenerator/web/util/TestGeneratorObjectMapper.java
```diff
@@ -0,0 +1,30 @@
+package eu.execom.testgenerator.web.util;
+
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Own implementation of {@link ObjectMapper}.
```
stanislavgrujic: Explain why do we need it. This JavaDoc does not say anything.

PR #160, opened 2018-06-28
Feature/auth/frontend auth implementation

sugrinov:

PR #174, opened 2018-07-06
Feature/auth/add auth configuration

sugrinov:
@stanislavgrujic Ignore javascript files since they were already reviewed #160.
Since these changes separate would break our development of other issues we decided to first create a pull request for frontend branch into backend and then create a pull request to develop.


web/src/main/java/eu/execom/testgenerator/web/controller/LoginController.java
```diff
@@ -0,0 +1,82 @@
...
+      return new ResponseEntity(HttpStatus.UNAUTHORIZED);
+    }
+
+    User user = userService.findByEmail(email);
+    if (user == null) {
+      user = new User();
+      user.setEmail(email);
+      user.setName(name);
+      userService.save(user);
+    }
```
stanislavgrujic: This logic belongs to `UserServiceImpl`, not in the controller. The controller should just be used for making calls to the service. Lines 68 to 74 should go to `userService.findByEmail()` with note that you should call save on the `userRepository`, not service method.

web/src/main/java/eu/execom/testgenerator/web/controller/LoginController.java
```diff
@@ -0,0 +1,82 @@
...
+  @RequestMapping(method = RequestMethod.POST, value = "/api/login/google")
+  public ResponseEntity googleLogin(@RequestBody String accessToken) throws IOException, GeneralSecurityException {
+    GoogleIdTokenVerifier verifier = new GoogleIdTokenVerifier
+      .Builder(new NetHttpTransport(), JacksonFactory.getDefaultInstance())
+      .setIssuer("accounts.google.com")
+      .build();
+
+    GoogleIdToken idToken = verifier.verify(accessToken);
+
+    return handleGoogleLogin(idToken);
```
stanislavgrujic: Handle these two exceptions in the method, add custom security exception and handle it in the GlobalExceptionHandler properly.

web/src/main/java/eu/execom/testgenerator/web/security/AuthenticationFilter.java
```diff
@@ -0,0 +1,64 @@
...
+  private TokenStore tokenStore;
+
+  public AuthenticationFilter(AuthenticationManager authenticationManager, TokenStore tokenStore) {
+    this.authenticationManager = authenticationManager;
+    this.tokenStore = tokenStore;
+  }
+
+  @Override
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+    throws IOException, ServletException {
```
stanislavgrujic: Handle these two exceptions in the method, add custom security exception and handle it in the GlobalExceptionHandler properly.

sugrinov: Since we are overriding a method from `GenericFilterBean` which throws those two exceptions,
I cannot make a more specific exception since those are two separate exceptions. Only thing I could do is extend one of the exceptions but that could lead to misinformation (regardless of which is thrown the stack trace would point to the extended exception). 

Could you provide some advice on how this could be handled?

stanislavgrujic: Leave it as is, then, just handle those two exceptions in the GlobalExceptionHandler.

web/src/main/java/eu/execom/testgenerator/web/controller/LoginController.java
```diff
@@ -0,0 +1,85 @@
...
+  private GoogleIdToken verifyGoogleToken(String accessToken) throws VerifyTokenException {
+    GoogleIdTokenVerifier verifier = new GoogleIdTokenVerifier
+      .Builder(new NetHttpTransport(), JacksonFactory.getDefaultInstance())
+      .setIssuer("accounts.google.com")
+      .build();
+
+    try {
+      return verifier.verify(accessToken);
+    } catch (GeneralSecurityException | IOException e) {
+      throw new VerifyTokenException("Token verification failed", e);
```
stanislavgrujic: Please rename this exception to `TokenVerificationFailedException`, it is more expressive.

web/src/main/java/eu/execom/testgenerator/web/util/VerifyTokenException.java
```diff
@@ -0,0 +1,8 @@
+package eu.execom.testgenerator.web.util;
+
+public class VerifyTokenException extends Exception  {
```
stanislavgrujic: If you extend from `RuntimeException`, you will not have to explicitly state that method throws it, which makes the code cleaner.

web/src/main/java/eu/execom/testgenerator/web/util/VerifyTokenException.java
```diff
@@ -0,0 +1,8 @@
...
+
+  public VerifyTokenException(String message, Throwable throwable) {
```
stanislavgrujic: Do not use `Throwable`, since `Error` also extends from it, and should not be used. Use `Exception`, instead.

PR #184, opened 2018-07-10
Error handling when backend returns empty array

vrackovic:
- Created component that renders passed error message
- Added check that returns normal component or error component if fetched data is empty array
- Changed style and responsive design


iscekic:
Please resolve conflicts and I'll review. 😄

web/src/main/frontend/src/base/util.js
```diff
@@ -48,3 +48,11 @@ export const questionFormNormalizer = questions => {
     category: value
   };
 };
+
+export const checkResponseData = responseData => {
+  if (responseData && responseData.length < 1) {
```
iscekic: I dislike this. We can be sure that the server will always return an array from endpoints where we use `checkResponseData` (@sugrinov right?).

So we can, instead of polluting all our api calls with `checkResponseData`, simply rely on the length of the array? Let me know what you think. @vrackovic

iscekic: All no data checks can then be as simple as `arr.length ? ... : ...`

web/src/main/frontend/src/pages/Formula.js
```diff
@@ -28,24 +29,43 @@ class Formula extends React.Component {
...
+        .$('recipe')
+        .map(category => <FormulaItem key={category.id} recipe={category} />)
+    ) : (
+      <NoDataAvailable
+        cssClass="formula_category__no_data"
+        message="Create some categories first"
+      />
+    );
+
+    // const categories = this.props.formulaForm
```
iscekic: Remove comment.

web/src/main/frontend/src/base/rest/categoryApi.js
```diff
@@ -1,8 +1,11 @@
 import {get, post, put, del, withLoader} from './requests';
+import {checkResponseData} from '../util';
 
 class CategoryApi {
   getAllCategories() {
-    return withLoader(get('/categories')).then(res => res.categoryResponses);
+    return withLoader(get('/categories')).then(res => {
```
sugrinov: Can't syntax stay the same as it was? Why do we need to wrap it with `{}`?

web/src/main/frontend/src/base/util.js
sugrinov: Correct, except if someone tries to access the page with no authentication in which case 401 will be returned with no data, but I don't know how it will be handled

iscekic: It will be caught by the error handler? Perhaps we can use our wrapper to handle 401s.

web/src/main/frontend/src/components/NoDataAvailable.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const NoDataAvailable = props => {
```
sugrinov: We don't need to use this syntax when a `Component` simply returns jsx. You can omit `{}` and `return()`

web/src/main/frontend/src/components/PanelControls.js
```diff
@@ -5,7 +5,7 @@ import Button from './ui/Button';
 class PanelControls extends React.Component {
   render() {
     return (
-      <div className={`col-12 panel__content ${this.props.containerClass}`}>
+      <div className={`panel__content ${this.props.containerClass}`}>
```
sugrinov: What if containerClass is `undefined`? Can it be undefined? :laughing:

sugrinov: Also, maybe we could use `classNames` for cases like this. I don't know if it's too late to add it now since we already have, I guess, pretty much all classes defined. @iscekic What do you think?

iscekic: It probably can, but it won't do anything noticable unless we have a `undefined` css class - in which case 💥

web/src/main/frontend/src/components/QuestionFormBody.js
```diff
@@ -41,7 +41,7 @@ class QuestionFormBody extends React.Component {
                 <div className="col-12 col-xl-3 text-left align-items-center d-flex">
                   <label className="label-inverted">Category</label>
                 </div>
-                <div className="col-12 col-xl-9 text-left input-inverted">
+                <div className="col-12 col-xl-9 text-left ">
```
sugrinov: `"col-12 col-xl-9 text-left "` You have an extra space :alien:

web/src/main/frontend/src/components/PanelControls.js
vrackovic: Yup, it can be `undefined`. Should i then use `classNames`? There is few more cases where classes are handled like this, so i purpose fixing this in new PR dedicated just for `classNames` implementation.

web/src/main/frontend/src/base/rest/dashboardApi.js
```diff
@@ -1,8 +1,11 @@
 import {get, withLoader} from './requests';
+import {checkResponseData} from '../util';
 
 class DashboardApi {
   getDashboardData() {
-    return withLoader(get('/dashboard')).then(res => res.dashboardTestDtos);
+    return withLoader(get('/dashboard')).then(res => {
```
sugrinov: Please replace this with just a call to `checkResponseData(res.categoryResponses)`

web/src/main/frontend/src/base/rest/formulaApi.js
```diff
@@ -1,8 +1,11 @@
 import {get, post, put, del, withLoader} from './requests';
+import {checkResponseData} from '../util';
 
 class FormulaApi {
   getAllFormulas() {
-    return withLoader(get('/formulas')).then(res => res.formulas);
+    return withLoader(get('/formulas')).then(res => {
```
sugrinov: Also here.

web/src/main/frontend/src/base/rest/questionApi.js
```diff
@@ -1,8 +1,11 @@
 import {get, post, put, del, withLoader} from './requests';
+import {checkResponseData} from '../util';
 
 class QuestionApi {
   getAllQuestions() {
-    return withLoader(get('/questions')).then(res => res.questions);
+    return withLoader(get('/questions')).then(res => {
```
sugrinov: Also here.

web/src/main/frontend/src/components/PanelControls.js
sugrinov: Yeah we can create an issue for this and fix it later. I hope we won't have `undefined` classes till then :see_no_evil:

web/src/main/frontend/src/pages/Formula.js
```diff
@@ -28,24 +29,38 @@ class Formula extends React.Component {
   };
 
   render() {
-    const categories = this.props.formulaForm
-      .$('recipe')
-      .map(category => <FormulaItem key={category.id} recipe={category} />);
+    const categories = !this.props.category.categories.noData ? (
```
iscekic: Redundant check?

vrackovic: Forgot to fix that one :laughing:

PR #213, opened 2018-07-19
Moving review test to dashboard

vrackovic:
- `ReviewTest` is activated by pressing `Edit` icon in dashboard.
- It is opened as modal
- Component is now loading real `Test` data from backend


web/src/main/frontend/src/base/stores/CompletedTestStore.js
```diff
@@ -1,74 +1,9 @@
...
-      type: 'CodeChallenge',
-      question: 'Hello fellow coder, please write us some clean code here:',
-      solution:
-        'if (true) {// give me points please, I need this job to support my family \n }',
-      availablePoints: 40
-    }
-  ];
+  @observable test;
 
+  @observable activateReview = false;
```
iscekic: Rename to `reviewActive`.

web/src/main/frontend/src/components/Test.js
```diff
@@ -59,7 +59,8 @@ class Test extends React.Component {
         <div className="review__container__header row-eq-height">
           <div className="review__container__counter row-eq-height d-flex justify-content-center ">
             <span>
-              Step {this.counter || 1} of {this.props.getQuestions().length}
+              Step {this.counter || 1} of
+              {this.props.questions && this.props.questions.length}
```
iscekic: When is `questions` falsy?

web/src/main/frontend/src/pages/TakeTest.js
```diff
@@ -13,7 +13,9 @@ class TakeTest extends React.Component {
   }
 
   questionsToReview = () =>
-    this.props.newTest.test.questions.map(getComponentByQuestionType);
+    this.props.newTest.test.questions.map(question =>
```
iscekic: Equivalent:

```js
this.props.newTest.test.questions.map(getComponentByQuestionType)
```

web/src/main/frontend/src/components/ReviewTest.js
```diff
@@ -0,0 +1,55 @@
...
+import CSSTransition from 'react-transition-group/CSSTransition';
+import {inject, observer} from 'mobx-react';
+import Test from './Test';
+import getComponentByQuestionType from './utility/getComponentByQuestionType';
+
+@inject('confirmPopup', 'completedTest')
+@observer
+class ReviewTest extends React.Component {
+  render() {
+    const questionsToReview = this.props.completedTest.test.questions.map(
```
iscekic: Can this be computed?

web/src/main/frontend/src/components/ReviewTest.js
```diff
@@ -0,0 +1,55 @@
...
+      question => getComponentByQuestionType(question, true)
+    );
+
+    return (
+      <div>
```
iscekic: Do we need to wrap with `div`?

web/src/main/frontend/src/components/Test.js
vrackovic: This was an issue when function has been triggered before data was loaded. It is now fixed and `questions` will never be falsy so check will be removed.

web/src/main/frontend/src/components/ReviewTest.js
vrackovic: Nope, we don't

web/src/main/frontend/src/base/stores/CompletedTestStore.js
```diff
@@ -83,6 +18,22 @@ class CompletedTestStore {
...
+    this.reviewActive = true;
+  }
+
+  @action.bound
+  exitReview() {
+    this.reviewActive = false;
+  }
+
+  @computed
+  get testQuestions() {
```
iscekic: Will there be some computation here in the future?

web/src/main/frontend/src/components/CodeChallengeQuestion.js
```diff
@@ -61,7 +61,7 @@ class CodeChallengeQuestion extends Component {
             value={this.state.value}
             onBeforeChange={this.updateCode}
             options={this.getOptions()}
-            mode={this.options.mode}
+            mode={this.state.mode}
```
iscekic: `state.mode`? Where is this defined?

web/src/main/frontend/src/components/ReviewTest.js
```diff
@@ -0,0 +1,53 @@
...
+import CSSTransition from 'react-transition-group/CSSTransition';
+import {inject, observer} from 'mobx-react';
+import Test from './Test';
+import getComponentByQuestionType from './utility/getComponentByQuestionType';
+
+@inject('confirmPopup', 'completedTest')
+@observer
+class ReviewTest extends React.Component {
+  render() {
+    const questionsToReview = this.props.completedTest.testQuestions.map(
```
iscekic: Can this be `computed`?

vrackovic: I can map trough `testQuestions` and return `question => getComponentByQuestionType(question, true)` from store, but question is is it smart to move jsx elements as computed in store? `getComponentByQuestionType` returns element render that depends of `question.type`

web/src/main/frontend/src/components/CodeChallengeQuestion.js
vrackovic: This field should say `this.state.syntax`. It was called `mode` before renamed to `syntax`.

web/src/main/frontend/src/components/ReviewTest.js
iscekic: Hm, you're right. Let's leave this here then.

web/src/main/frontend/src/base/stores/CompletedTestStore.js
vrackovic: Same as comment about returning `JSX` from store, that is only place that calls `testQuestion`

PR #222, opened 2018-07-23
Feature/sugrinov add applicant to token store

sugrinov:


web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -102,4 +115,23 @@
     testService.submit(submitTestRequest.getTestId(), convertedAnswers);
     return new ResponseEntity<>(HttpStatus.OK);
   }
+
+  private ResponseEntity<TestDto> checkUserHasTestAccess(HttpServletRequest httpRequest, Test test) {
+    Optional<String> optionalToken = Optional.ofNullable(httpRequest.getHeader(APPLICANT_TOKEN_KEY));
+
+    boolean tokenNotPresent = !(optionalToken.isPresent() && tokenStore.containsKey(optionalToken.get()));
+    if (tokenNotPresent) {
+      return new ResponseEntity<>(HttpStatus.BAD_REQUEST); // 400
```
luinnarn: Just a suggestion, one possible situation is that a token is present, but not contained in tokenStore, so perhaps the boolean should be named something like `validTokenNotPresent` or `validTokenNotFound`?

sugrinov: Yes that is a valid suggestion, I will change the name. But both of those imply(to me atleast) that token is present in header but it is not present in token store and it doesn't have to be present in the header therefore we have the check. What do you think about `isTokenPresentAndValid`? Maybe it is an overkill but makes most sense to me :laughing:

luinnarn: `isTokenPresentAndValid` is OK but then you would have to negate it, `if (!isTokenPresentAndValid)`, `if (validTokenNotPresent)` simply seems more readable.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -41,6 +48,8 @@
   private FormulaService formulaService;
   @Autowired
   private UserService userService;
+  @Autowired
```
dujfalusi: Please fix formatting

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -102,4 +115,23 @@
     testService.submit(submitTestRequest.getTestId(), convertedAnswers);
     return new ResponseEntity<>(HttpStatus.OK);
   }
+
+  private ResponseEntity<TestDto> checkUserHasTestAccess(HttpServletRequest httpRequest, Test test) {
```
dujfalusi: Shouldn't this, be a part of some user service?

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
sugrinov: What should be formatted? You mean spaces between services?

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
sugrinov: Most of the logic here is to check which `Response` should be returned so it makes more sense to be here.

web/src/test/java/eu/execom/testgenerator/web/controller/TestControllerTest.java
```diff
@@ -92,17 +96,24 @@ public void shouldCreateNewTest() throws Exception {
     Test test = new Test();
     test.setId(5L);
 
+    String token = "token";
+
+
```
dujfalusi: Please remove this unnecessary empty line.

PR #225, opened 2018-07-25
Feature/sugrinov fe fix formula update

sugrinov:


web/src/main/frontend/src/components/forms/FormulaForm.js
```diff
@@ -8,33 +8,41 @@ const fields = [
...
 
 const hooks = {
   onSuccess(form) {
     let formValues = form.values();
-    const recipe = formValues.recipe;
+    const recipes = formValues.recipes.map(recipe => {
+      return {
+        numberOfQuestions: recipe.numberOfQuestions,
+        categoryId: recipe.category.id.value
+          ? recipe.category.id.value
```
iscekic: Why do we only sometimes have `value`?

web/src/main/frontend/src/pages/Formula.js
```diff
@@ -20,7 +20,10 @@ class Formula extends React.Component {
   };
 
   handleRowClick = selectedFormula => {
-    this.props.formulaForm.update(selectedFormula);
+    const selected = this.props.formula.formulas.find(
```
iscekic: Can this ever be `undefined`?

web/src/main/frontend/src/components/forms/FormulaForm.js
sugrinov: Because react-select returns an object which has `{ value, label }` and backend returns a category object `{ id, name, category }`. When we are editing a formula with existing recipe, it will contain `category` and when we add a new recipe to that it will contain an object from `react-select` as well.

web/src/main/frontend/src/pages/Formula.js
sugrinov: Shouldn't happen since this happens `onClick` on table row where all formulas are listed and they have their id :laughing:

web/src/main/frontend/src/components/forms/FormulaForm.js
iscekic: Leave a comment please. 😄

sugrinov: I guess I can have a look at `react-select` in a bit more detail and try to implement some wrapper that always only returns a value when we select an item instead of returning an object since we have no use for label from it (at least i think we don't). But it would be in another PR.

sugrinov: Will leave it :space_invader:

PR #229, opened 2018-07-25
Disallow applicant from accessing tests that are not his

sugrinov:


web/src/main/frontend/src/base/rest/testApi.js
```diff
@@ -1,8 +1,12 @@
...
+import {get, del, withLoader, withGuestToken} from './requests';
+import {
+  GUEST_TOKEN_KEY,
+  getGuestToken
+} from '../../components/utility/httpUtil';
 
 class testApi {
   getTest(testId) {
-    return withLoader(get(`/tests/${testId}`));
+    return withLoader(withGuestToken(get(`/tests/${testId}`)));
```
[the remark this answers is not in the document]
sugrinov: Yes, we need to check if test with that id belogs to current applicant

web/src/main/frontend/src/base/rest/requests.js
```diff
@@ -1,24 +1,34 @@
...
-  request.delete(getLink(url)).set('X-Auth-Token', getToken());
+  request.delete(getLink(url)).set(AUTH_TOKEN_KEY, getAuthToken());
 
 export const withLoader = req => {
   LoaderStore.incrementLoader();
   return req
     .then(res => {
       LoaderStore.decrementLoader();
+      if (res.headers['x-guest-token']) {
+        return res;
```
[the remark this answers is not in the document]
sugrinov: This is a workaround for passing headers to `testApi`, since current implementation of `withLoader` only returns response body, and in that case we need headers also.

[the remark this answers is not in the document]
iscekic: Let's make `withLoader` return the entire response and have `get`, `post`, etc. extract the body.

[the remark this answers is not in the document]
sugrinov: Sure, I will make a new PR.

web/src/main/frontend/src/components/utility/httpUtil.js
```diff
@@ -0,0 +1,9 @@
+export const AUTH_TOKEN_KEY = 'X-Auth-Token';
+export const GUEST_TOKEN_KEY = 'X-Guest-Token';
```
iscekic: Why not `const GUEST_TOKEN_KEY = 'x-guest-token'`? HTTP headers are case-insensitive anyway. 😄

iscekic: Reuse token keys for session storage.

web/src/main/frontend/src/base/rest/requests.js
```diff
@@ -1,32 +1,44 @@
...
+} from '../../components/utility/httpUtil';
 
 export const get = url =>
   request
     .get(getLink(url))
-    .set('X-Auth-Token', getToken())
+    .set(AUTH_TOKEN_KEY, getAuthToken())
+    .then(res => res.body);
+
+export const getWithHeaders = (url, headers) =>
```
iscekic: Move `getWithHeaders` and `rawPost` to separate file.

web/src/main/frontend/src/base/rest/requests.js
```diff
@@ -1,32 +1,30 @@
 import request from 'superagent';
 import {getLink} from '../getLink';
 import LoaderStore from '../stores/LoaderStore';
+import {AUTH_TOKEN_KEY, getAuthToken} from '../../components/utility/httpUtil';
 
 export const get = url =>
   request
     .get(getLink(url))
-    .set('X-Auth-Token', getToken())
+    .set(AUTH_TOKEN_KEY, getAuthToken())
```
iscekic: Can we shorten this to `.set(authApi.token)`?

web/src/main/frontend/src/base/rest/testApi.js
```diff
@@ -1,13 +1,21 @@
...
+import {del, withLoader} from './requests';
+import {
+  getGuestToken,
+  GUEST_TOKEN_KEY
+} from '../../components/utility/httpUtil';
 
 class testApi {
   getTest(testId) {
-    return withLoader(get(`/tests/${testId}`));
+    return withLoader(request.get(`/tests/${testId}`).set(testApi.token()));
```
iscekic: Shouldn't this be `.set(testApi.token)`?

web/src/main/frontend/src/base/rest/requests.js
iscekic: Actually, it might create a circular dependency. Just leave it as is if that's the case. 😄

sugrinov: Yup, tried it and it wasn't working :man_juggling:

PR #237, opened 2018-07-27
Returned only body from server response in every superagent request

sugrinov:
* Added postWithHeaders method that will be used in requests that require headers


web/src/main/frontend/src/base/rest/requests.js
```diff
@@ -3,25 +3,37 @@ import {getLink} from '../getLink';
...
+    .set('X-Auth-Token', getToken())
+    .then(res => res.body);
 
 export const post = (url, data) =>
+  request
+    .post(getLink(url), data)
+    .set('X-Auth-Token', getToken())
+    .then(res => res.body);
+
+export const postWithHeaders = (url, data) =>
```
iscekic: Do you think `rawPost` would be a better name?

sugrinov: I think it would not be immediately clear what it does and I would have to look at the other methods to understand the logic behind it, but then again I can't think of a name that would suit it :confused: .

iscekic: I think it's relatively clear:

`post - posts, sets token, extracts body`
`rawPost - posts, sets token`

What do you think?

sugrinov: Sure, will change it :+1:

PR #244, opened 2018-07-31
Use getOne instead of findOne in formulaService update method

sugrinov:
* Changed findOne to getOne so we get EntityNotExists exception
   if formula is not found


core/src/main/java/eu/execom/testgenerator/core/service/FormulaServiceImpl.java
```diff
@@ -64,7 +64,7 @@ public Formula findByName(String name) {
   public Formula update(Formula updatedFormula) {
     checkIfFormulaExists(updatedFormula);
 
-    Formula oldFormula = formulaRepository.findOne(updatedFormula.getId());
+    Formula oldFormula = formulaRepository.getOne(updatedFormula.getId());
```
stanislavgrujic: Why did you make this change?

stanislavgrujic: Disregard my comment... :)

PR #248, opened 2018-07-31
Dropdown overflow rendering

vrackovic:
- Fixed rendering of opened dropdown outside if parent component


iscekic:
Let's upgrade `react-select` instead.

Issue #196, opened 2018-07-12
Implement functionality for assigning reviewer to test on Dashboard


sugrinov:
When clicking `Edit` button on a test in Dashboard we should be given an option to assign a reviewer to that test

vrackovic:
Layout and mock done in #226

millevlada:
actually, reviewers are assigned by doing 'assign reviewer...' action (first icon), and that is implemented with fx #278 . It is not clear what 'edit test' icon should do?

millevlada:
Ok, I presume that 'edit test' is actually 'do review'... This would be done with other user story - #290, so I'll close this one.

PR #249, opened 2018-07-31
Assigning reviewer to test - form and submit

vrackovic:
- Added assignReviewerForm
- Form submits data in format `{reviewerId: {int}, testId: {int}}`
- Data is submitted on `/test/assignReviewer`

Resolving issue #196


web/src/main/frontend/src/base/rest/dashboardApi.js
```diff
@@ -1,9 +1,22 @@
-import {get, withLoader} from './requests';
+import {get, post, withLoader} from './requests';
 
 class DashboardApi {
   getDashboardData() {
     return withLoader(get('/dashboard')).then(res => res.dashboardTestDtos);
   }
+
+  getReviewers() {
+    return withLoader(get('/users/reviewers')).then(res =>
```
iscekic: `.then(this.prepareReviewers)`

web/src/main/frontend/src/base/stores/DashboardStore.js
```diff
@@ -34,13 +36,33 @@ class DashboardStore {
...
+      .getReviewers()
+      .then(reviewers => (this.reviewers = reviewers));
+  }
+
+  @action.bound
+  assignReviewer(reviewer) {
+    const testWithReviewer = {...reviewer, testId: this.testToAssign};
+    return dashboardApi
+      .submitAssignReviewer(testWithReviewer)
+      .then(res => res && this.disableReviewerPopup());
```
iscekic: Why are we checking `res` truthy-ness here?

vrackovic: Idea was to check for `res.status` in case something goes wrong, but it will be removed.

PR #252, opened 2018-08-02
Create endpoint for assigning reviewer to the selected test

sugrinov:
* Added a service method in test for assigning a reviewer to a given test
  * Created unit and integration test methods
  * Added endpoint for assigning reviewer
  * Added antMatcher that accepts only PUT method to '/assign-reviewer'
   route and requires user to be authenticated


web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -120,6 +122,14 @@
     return new ResponseEntity<>(HttpStatus.OK);
   }
 
+  @PutMapping(value = "/assign-reviewer")
+  public ResponseEntity assignReviewer(@RequestBody AssignReviewerRequest assignReviewer) {
+    Test test = testService.findOne(assignReviewer.getTestId());
```
stanislavgrujic: This should be part of `assignReviewer` method, and it should have `testId` as a parameter.

web/src/test/java/eu/execom/testgenerator/web/integration/TestControllerIntegrationTest.java
```diff
@@ -157,4 +161,24 @@ private void createTestThroughController(Formula formula, User applicant) throws
                                            .contentType(MediaType.APPLICATION_JSON_UTF8_VALUE)).andReturn();
   }
 
+  @org.junit.Test
+  public void shouldAssignReviewer() throws Exception {
+    // Given
+    Test test = TestMother.test(1L, formula, applicant);
+    Test updatedTest = testService.update(test);
```
stanislavgrujic: Why use update method, and not save, as is in line 171? I assume they work the same, but it's not intuitive when reading the code.

sugrinov: I agree that it's not intuitive but `TestService` doesn't have a `save` method since there was no need for it. Currently the only place where we would need a save method is in tests.

PR #253, opened 2018-08-02
Upgrading react select to v2

vrackovic:
- Migrated react-select version 1.3 to v2
- Fixed conflicts caused by migration


web/src/main/frontend/src/base/util.js
```diff
@@ -42,14 +42,16 @@ export const prepareTimestampForView = t =>
...
     return questions;
   }
 
   const {category: {value}} = questions;
+  const type = questions.type.value;
 
   return {
     ...questions,
-    category: value
+    category: value,
```
iscekic: Rename `value` to `category` so we can have:

```js
{
  ...questions,
  category,
  value
}
```

web/src/main/frontend/src/components/QuestionFormBody.js
```diff
@@ -8,20 +8,23 @@ import questionTypes, {
...
 
-  handleCategorySelect = selectedCategory => {
-    this.props.questionForm.update({category: selectedCategory.value});
+  handleTypeChange = selectedType => {
+    this.props.questionForm.$('type').set(selectedType);
   };
 
   showAnswers = () => {
-    return this.props.questionForm.$('type').value ===
+    // TODO: Change value.value into something more readable
```
iscekic: Can we solve this as part of this PR? In a normalizer or something.. 😄

web/src/main/frontend/src/components/utility/checkForSelectValue.js
```diff
@@ -0,0 +1,4 @@
+const checkForSelectValue = option =>
```
iscekic: What happens if we delete this? 😄

iscekic: Let's try and refactor so that form reset works properly and we can remove this.

PR #256, opened 2018-08-06
Dashboard pagination

vrackovic:
- Created `Pagination` component
- Installed `lodash.debounce`
- Implemented debounce to load page after 350ms after last user input
- `first`, `previous`, `next` and `last` buttons implemented for pagination


web/src/main/frontend/src/base/rest/dashboardApi.js
```diff
@@ -1,8 +1,16 @@
...
 class DashboardApi {
-  getDashboardData() {
-    return withLoader(get('/dashboard')).then(res => res.dashboardTestDtos);
+  getDashboardData(currentPage) {
+    return withLoader(
+      request
+        .get(`api/dashboard?pageNum=${currentPage}`)
+        .then(res => ({
+          data: res.body.dashboardTestDtos,
+          totalPages: res.headers.totalpages
```
iscekic: Total pages in header? @sugrinov What blasphemy is this?

web/src/main/frontend/src/components/Pagination.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import Button from './ui/Button';
+
+class Pagination extends React.Component {
+  handleChange = e => {
+    const val = e.target.value;
+    this.props.changeCurrentPage(val);
+  };
+  render() {
```
iscekic: `\n`

web/src/main/frontend/src/components/Pagination.js
```diff
@@ -0,0 +1,47 @@
...
+    return (
+      <div className="dashboard__pagination_wrapper">
+        <Button
+          title="<<"
+          cssClass="dashboard__pagination_button"
+          click={() => this.props.goToPage('first')}
```
iscekic: Why use a string `'first'` instead of a number `0`?

web/src/main/frontend/src/pages/Dashboard.js
```diff
@@ -17,6 +19,37 @@ class Dashboard extends Component {
     this.props.dashboard.fetchDashboardData();
   }
 
+  goToPage = control => {
```
iscekic: I dislike this. Lets have the `Pagination` component accept a function which sets the page (does the work), and leave all the pagination logic inside the component.

web/src/main/frontend/src/base/rest/dashboardApi.js
sugrinov: This was implemented before I started working on backend :see_no_evil:

web/src/main/frontend/src/components/Pagination.js
vrackovic: No particular reason. Will be removed :smile:

web/src/main/frontend/src/base/rest/dashboardApi.js
```diff
@@ -1,8 +1,14 @@
...
+import request from 'superagent';
 
 class DashboardApi {
-  getDashboardData() {
-    return withLoader(get('/dashboard')).then(res => res.dashboardTestDtos);
+  getDashboardData(currentPage) {
+    return withLoader(
+      request.get(`api/dashboard?pageNum=${currentPage}`).then(res => ({
+        data: res.body.dashboardTestDtos,
+        totalPages: res.headers.totalpages
```
iscekic: Leave a `TODO` here for @sugrinov 😄

web/src/main/frontend/src/pages/Dashboard.js
```diff
@@ -17,6 +19,15 @@ class Dashboard extends Component {
     this.props.dashboard.fetchDashboardData();
   }
 
+  handleCurrentPage = value => {
+    const pageDisplay = Number(value);
```
iscekic: Do we need a `NaN` check here?

web/src/main/frontend/src/components/Pagination.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import Button from './ui/Button';
+import {observer, inject} from 'mobx-react';
+
+@inject('dashboard')
```
iscekic: Let's make this component reusable - the parent should pass appropriate click handlers and this component just uses them.

web/src/main/frontend/src/pages/Dashboard.js
vrackovic: Don't think we need it because `shouldChangePage` will be `false` in all cases except entered value is between 0 and `totalPages`

web/src/main/frontend/src/pages/Dashboard.js
```diff
@@ -18,6 +20,13 @@ class Dashboard extends Component {
...
+    return this.props.dashboard.setCurrentPage(
+      !isNaN(pageDisplay) && pageDisplay
```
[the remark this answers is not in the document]
vrackovic: Yup, `this.props.dashboard.setCurrentPage(false);` is called in case when `NaN` is passed as value. 
Should i check and early return if value passed is `NaN`. That way `setCurrentPage()` will not be called?

web/src/main/frontend/src/components/Pagination.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import Button from './ui/Button';
+
+class Pagination extends React.Component {
+  state = {
+    currentPage: this.props.currentPage || 1
```
iscekic: Don't use local state, we're duplicating data we already have in the store.

millevlada:
Obsoleted with #305

PR #257, opened 2018-08-06
Feature/be/sugrinov/add test timer

sugrinov:


core/src/main/java/eu/execom/testgenerator/core/service/ScheduledTasks.java
```diff
@@ -0,0 +1,55 @@
...
+   * Checks if there are tests for which the time has expired.
+   * This check happens every 5 minutes.
+   */
+  @Scheduled(fixedDelay = FIVE_MINUTES)
+  public void checkIsTestTimeExpired() {
+    List<Test> notCompletedTests = testService.findAllNotCompleted();
+    notCompletedTests.forEach(this::calculateTestTimePassed);
+  }
+
+  private void calculateTestTimePassed(Test test) {
```
stanislavgrujic: Naming is not exactly correct, since it calculates the time passed since the test was started, but it also completes expired tests, so it would be more meaningful to name the method as `handleExpiredTests`. Also, even more readable would be to make this code more functional so that it is written something like
```
notCompletedTests.stream().filter(expiredTests()).forEach(this::handleExpiredTest);

// expiredTests() would check if the test is expired and have logic in lines 39 - 43 + return result of line 50
// handleExpiredTest would set the test to completed, which are basically lines 51 and 52
```
This would ensure that each method has exactly one responsibility, instead of having side effects of each of them in another method.

core/src/main/java/eu/execom/testgenerator/core/service/ScheduledTasks.java
```diff
@@ -0,0 +1,55 @@
...
+  public ScheduledTasks(TestService testService) {
+    this.testService = testService;
+  }
+
+  /**
+   * Checks if there are tests for which the time has expired.
+   * This check happens every 5 minutes.
+   */
+  @Scheduled(fixedDelay = FIVE_MINUTES)
+  public void checkIsTestTimeExpired() {
```
stanislavgrujic: Rather name it `findAndCompleteExpiredTests()`

core/src/main/java/eu/execom/testgenerator/core/service/ScheduledTasks.java
```diff
@@ -0,0 +1,55 @@
...
+    List<Test> notCompletedTests = testService.findAllNotCompleted();
```
stanislavgrujic: It would be more expressive if you name the variable `activeTests`.

core/src/main/java/eu/execom/testgenerator/core/service/ScheduledTasks.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.util.List;
+
+/**
+ * Schedule tasks for execution after certain period of time.
+ */
+@Component
+public class ScheduledTasks {
```
stanislavgrujic: The name is way to generic. Please rename it to depict what it exactly does and do not combine other scheduled tasks (in the future) with this one, since it would break the [single responsibility principle](url
https://en.wikipedia.org/wiki/Single_responsibility_principle).

core/src/test/java/eu/execom/testgenerator/core/service/ScheduledTasksTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+@RunWith(MockitoJUnitRunner.class)
+public class ScheduledTasksTest {
+
+  @Mock
+  private TestServiceImpl testService;
+
+  @org.junit.Test
+  public void shouldCompleteExpiredTests() {
+    ScheduledTasks scheduledTasks = new ScheduledTasks(testService);
```
stanislavgrujic: You should not manually create it, but let Mockito do that for you by creating a field
```
@InjectMocks
private ScheduledTasks scheduledTasks;
```

core/src/main/java/eu/execom/testgenerator/core/service/ScheduledTasks.java
```diff
@@ -0,0 +1,55 @@
...
+  @Scheduled(fixedDelay = FIVE_MINUTES)
+  public void checkIsTestTimeExpired() {
+    List<Test> notCompletedTests = testService.findAllNotCompleted();
+    notCompletedTests.forEach(this::calculateTestTimePassed);
+  }
+
+  private void calculateTestTimePassed(Test test) {
+
+    LocalDateTime testTimeTaken = test.getTimeTaken();
+    LocalDateTime now = LocalDateTime.now();
```
stanislavgrujic: In general, this works, but it can create issues while testing. It would be convenient to introduce a simple class which you can inject in services where you need a time, and which you can mock in your unit tests. Class would be similar to:
```
@Component
public class TimeProvider {

  public LocalDateTime getTime() {
    return LocalDateTime.now();
  }
}

// and the usage would be
LocalDateTime now = timeProvider.getTime();

// and in tests you can then provide any time you want, instead of having to rely on current time by using something like:
given(timeProvider.getTim()e).willReturn(timeYouNeed);
```

core/src/main/java/eu/execom/testgenerator/core/service/ScheduleTestExpirationChecker.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.util.List;
+
+/**
+ * Schedule checks for test expiration.
+ */
+@Component
+public class ScheduleTestExpirationChecker {
```
stanislavgrujic: I would rename this to `ScheduledExpiredTestsHandler`.

Issue #180, opened 2018-07-09
Dashboard page Pagination


vrackovic:
- Dashboard page should have pagination implemented
- Multiple pages style should be implemented

millevlada:
Fixed with #305

PR #259, opened 2018-08-07
Dashboard query search

vrackovic:
- Dashboard loading searched parameter using `query` from `superagent`
- Previously implemented pagination works as expected with search
- `debounce` added on user input so that it fires request 500ms after user stopped typing

NOTE: Merge after #256 
Resolves issue #180


millevlada:
Obsoleted with #305

PR #277, opened 2018-08-22
Fix reviewer assignment after react-select update

sugrinov:


web/src/main/frontend/src/components/forms/AssignReviewerForm.js
```diff
@@ -14,7 +14,9 @@ const fields = {
 const hooks = {
   onSuccess(form) {
     const formValues = form.values();
-    DashboardStore.assignReviewer({reviewerId: formValues.reviewer.value});
+    DashboardStore.assignReviewer({reviewerId: formValues.reviewer.value}).then(
+      () => form.reset()
```
iscekic: `.then(form.reset)`

sugrinov: I tried it in another place and it doesn't want to cooperate because it needs an event.

iscekic: It's dangerous to go alone. Here, take [this](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Function/bind)!

sugrinov: What would be the difference? What improvements would we achieve by that?

iscekic: 1) Looks nicer (debatable)
2) I thought we would avoid creating an anonymous function, however `bind` also returns a function so no gain there. We could potentially reuse the bound reference, but that's overkill.

Reason 1 alone isn't enough to warrant a change, I'll approve. 😄

PR #279, opened 2018-08-22
Create a request for updating current test state

sugrinov:
* Change how multiple choice answers are handled
  * Add an onChange handler for multiple choice fields


web/src/main/frontend/src/base/rest/testApi.js
```diff
@@ -25,6 +25,14 @@ class testApi {
     );
   }
 
+  updateTest(testSolution) {
+    return request
+      .put(getLink('/tests/update-answers'))
+      .send(testSolution)
+      .set(testApi.token)
+      .then(res => res);
```
iscekic: Why do we need this?

web/src/main/frontend/src/base/util.js
```diff
@@ -73,7 +73,9 @@ export const checkQuestionType = (acc, question) => {
     case questionTypes.MULTIPLE_CHOICE:
       normalizedForm[questionId] = {
         questionType,
-        testQuestionAnswerIds: answer
+        testQuestionAnswerIds: Object.keys(multipleChoice).filter(
+          key => multipleChoice[key]
```
iscekic: Will this work if `multipleChoice[key]` is a falsy (but valid) value, like 0?

sugrinov: Yes, because the structure of it is an Object
```json
{
  "0": true,
  "1": false
}
```

PR #281, opened 2018-08-27
Create email service

sugrinov:
* Added endpoint for sending invitation mail to users so they can be added as reviewers
  * Added endpoint for accepting invitation


core/src/main/java/eu/execom/testgenerator/core/service/UserInviteServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+import eu.execom.testgenerator.core.model.UserInvite;
+import eu.execom.testgenerator.core.model.enumeration.UserRole;
+import eu.execom.testgenerator.core.repository.UserInviteRepository;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Slf4j
+@Service
+public class UserInviteServiceImpl implements UserInviteService {
```
stanislavgrujic: This class can be package private, no need to leak implementation details outside of the package.

core/src/main/java/eu/execom/testgenerator/core/service/EmailServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.mail.SimpleMailMessage;
+import org.springframework.mail.javamail.JavaMailSender;
+import org.springframework.stereotype.Component;
+
+import java.util.UUID;
+
+@Component
+public class EmailServiceImpl implements EmailService {
```
stanislavgrujic: This class can be package private, no need to leak implementation details outside of the package.

web/src/main/java/eu/execom/testgenerator/web/config/MailSenderConfig.java
```diff
@@ -0,0 +1,49 @@
...
+    mailSender.setHost(host);
+    mailSender.setPort(port);
+
+    mailSender.setUsername(username);
+    mailSender.setPassword(password);
+
+    javaMailProperties.put("mail.transport.protocol", "smtp");
+    javaMailProperties.put("mail.smtp.auth", auth);
+    javaMailProperties.put("mail.smtp.starttls.enable", ttlsEnable);
+    javaMailProperties.put("mail.debug", "true");
```
stanislavgrujic: Two hardcoded values should also be configurable via properties.

Issue #282, opened 2018-09-07
Allow user to have multiple roles


millevlada:
Currently each user can have single role. It causes strange rules like admin is also code reviewer. Since this is show case app as well, allow adding multiple roles to users to enable split of responsibilities of roles.

To achieve this, current model has to be refactored a bit, including needed data migration script. It is noticed that existing liquibase schema changes and migration scripts are written in less readable xml format, would be nicer if we could write them as regular SQL scripts. These could be reorganized a bit as well, to allow easier keeping track of order of changes.

Also, unit tests needs to be modified for change.

PR #283, opened 2018-09-07
[#282] Allow user to have multiple roles

millevlada:
Added SQL as preferred format of db changes scripts
Some code cleanup in unit tests


millevlada:
Ok, not all tests are fixed, still needs some work...

PR #293, opened 2018-09-12
Filter out not valid multiple choice question

mlcvijetinovic:
* Enrich test question with new method to avoid creating and
  adding to test non valid multiple choice questions
* Code refactoring


core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -26,6 +26,8 @@
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static java.util.stream.Collectors.toList;
```
dujfalusi: You have two imports of `Collectors` in line 75 you use `toList()` in line 94 you use `Collectors.toList()`

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -118,6 +108,44 @@ public Test create(Formula formula, User user) {
...
+      TestQuestion testQuestion = new TestQuestion();
+      testQuestion.setQuestion(question);
+      testQuestion.setTest(test);
+
+      List<TestQuestionAnswer> testQuestionAnswers = new ArrayList<>();
+
+      if (question.isMultipleChoice()) {
+        List<TestQuestionAnswer> multipleAnswerList = question.getAnswers()
+          .stream()
+          .map(mapTestQuestionAnswer(testQuestion))
```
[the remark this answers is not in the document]
mlcvijetinovic: Question are previous filtered and at this point we can have only active questions.

core/src/main/java/eu/execom/testgenerator/core/model/Question.java
```diff
@@ -86,4 +86,13 @@ public boolean isCodingChallenge() {
   public int getNumberOfCorrectAnswers() {
     return (int) answers.stream().filter(Answer::isCorrect).count();
   }
+
+  public boolean isValidQuestion() {
+
+    if (this.isMultipleChoice()) {
+      return !this.getAnswers().isEmpty() || this.getAnswers().stream().anyMatch(Answer::isActive);
```
millevlada: it is enough just: 
this.getAnswers().stream().anyMatch(Answer::isActive);

core/src/main/java/eu/execom/testgenerator/core/service/QuestionServiceImpl.java
```diff
@@ -43,7 +46,10 @@ public Question findOne(Long id) {
   public Question update(Question question) {
     Question oldQuestion = questionRepository.getOne(question.getId());
```
millevlada: Seems that this would always cause non necessary obsoleting and replicating questions and answers, even when there are no changes in question. It is not concern of this PR, so make separate enhancement task for this issue.

PR #297, opened 2018-09-12
Vmij 296 separation of controls

millevlada:


millevlada:
Agreed with @iscekic that later we can add improvement task to properly model this Table component to be customizable by parent wrapper components. For now it is just split into 2 separate components.

Issue #269, opened 2018-08-09
Move pagination in request body


sugrinov:

millevlada:
Fixed with #305

PR #305, opened 2018-09-18
Vmij 259 paging and search on dashboard

millevlada:
Merged together what was work in progress for #256, #259, and finished what was left (#269)  to make it work ;)


web/src/main/frontend/src/base/rest/dashboardApi.js
```diff
@@ -1,8 +1,19 @@
 import {get, put, post, withLoader} from './requests';
 
 class DashboardApi {
-  getDashboardData() {
-    return withLoader(get('/dashboard')).then(res => res.dashboardTestDtos);
+  getDashboardData(currentPage, queryString) {
+    let queryStringPart = '';
```
iscekic: Can use const & ternary here

iscekic: Also a simple truthy check should be enough?

web/src/main/frontend/src/base/stores/DashboardStore.js
```diff
@@ -120,6 +131,53 @@ class DashboardStore {
...
+      this.currentPage--;
+      this.fetchDashboardData();
+    }
+    return this.currentPage;
+  }
+
+  @action.bound
+  setQueryString(queryString) {
+    if (queryString === '') {
+      queryString = undefined;
```
iscekic: There should be no need for this

web/src/main/frontend/src/components/Pagination.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import Button from './ui/Button';
+
+class Pagination extends React.Component {
+  handleChange = e => {
+    const val = e.target.value;
```
iscekic: Can use destructuring

Issue #310, opened 2018-09-19
Problems with checkbox


millevlada:
Original bug that was reported is that in maintenance of questions we do not see what answer are marked as correct.
After research, it is noticed that Checkbox component is written badly (due to recent refactoring done by me), and it has to be fixed - by cleaning the code.

millevlada:
Closed with #311

PR #311, opened 2018-09-19
[#310] Cleanup of Checkbox control.

millevlada:
Fixing issue with not seeing correct answers in Question maintenance.
Also fixed issue with consequent reviews that were scoring results from previous tests ;)
Cleanup of some obsoleted code.


web/src/main/frontend/src/components/ui/Checkbox.js
```diff
@@ -1,27 +1,24 @@
...
-  let stylePart = '';
-  if (
-    (props.checked && !props.isCorrect) ||
-    (!props.checked && props.isCorrect)
-  ) {
-    stylePart = 'wrong';
-  } else if (props.checked && props.isCorrect) {
-    stylePart = 'correct';
-  }
+  let stylePart = props.labelStyle || '';
```
iscekic: This can be const

web/src/main/frontend/src/components/ui/Checkbox.js
```diff
@@ -1,27 +1,24 @@
...
 
   const answerTextClassName = `label_text ${stylePart}`;
 
+  let propsContainer = props.formField;
+
+  if (props.checked !== undefined) {
```
iscekic: These can be truthy checks

web/src/main/frontend/src/components/ReviewMultipleChoiceQuestion.js
```diff
@@ -5,17 +5,26 @@ import Checkbox from './ui/Checkbox';
 @observer
 class ReviewMultipleChoiceQuestion extends Component {
   options = this.props.options.map(option => {
+    const isChecked = this.props.answer.testQuestionAnswer.testQuestionAnswerIds.includes(
+      option.id
+    );
+    const isCorrect = this.props.answer.correctTestQuestionAnswerIds.includes(
+      option.id
+    );
+    let labelStyle = undefined;
```
iscekic: No need to explicitly set to `undefined`, `let labelStyle` is enough

PR #316, opened 2018-09-24
 Send notification mail to administrators

mlcvijetinovic:
- Add additional value check to avoid sending notification to administrators if there is no review changes.
 - Add new test and modify previous written test to work with new logic


web/src/test/java/eu/execom/testgenerator/web/controller/ReviewControllerTest.java
```diff
@@ -157,6 +165,37 @@ public void testApplyTestReview() throws Exception {
...
+
+    when(testService.findOne(1L)).thenReturn(test);
+
+    User loggedInUser = reviewer(3L);
+
+    SubmitReviewRequest submitReviewRequest = new SubmitReviewRequest();
+    submitReviewRequest.setTestId(1L);
+
+    List<User> administrators = new ArrayList<>();
+    administrators.add(admin(4L));
```
stanislavgrujic: Please use `List<User> administrators = Arrays.asLisT(admin(4L));`

PR #329, opened 2018-09-28
Fixed wrong email content when invitation mail is sent

mlcvijetinovic:
- Enrich path to invitation link works
- Fixed failing tests


core/src/main/java/eu/execom/testgenerator/core/util/Constants.java
```diff
@@ -4,7 +4,7 @@
 
   private Constants() { }
 
-  public static final String PATH_API_USERS = "/api/users";
+  public static final String PATH_API_USERS = "/oopt/api/users";
```
millevlada: use value from application properties

Issue #338, opened 2018-11-30
Imlement UI notifications


vrackovic:
- Remove current `MsgFlare` notification implementation
- Implement `react-tostify`, and support notification types

Issue #341, opened 2018-12-03
Support notification messages


vrackovic:
- [x] All api calls that trigger `catch`
- [x] Category submit, update and delete
- [x] Question submit, update and delete
- [x] Formula submit, update and delete
- [x] Invite employee popup
- [x] Assign reviewer
- [x] Review test
- [x] Delete test
- [x] Submit test

Issue #340, opened 2018-12-03
Add redirection after finishing test


vrackovic:
Current flow: 
- After test is done, `logout` button is shown

Expected flow:
- After user submits test, he's auto-redirected to welcome page, and 'successfully submitted' message is displayed

vrackovic:
Resolved with #349

PR #350, opened 2018-12-03
Replacing msg flare with react tostify

vrackovic:
- Installed `react-toastify` npm package
- Supported api call catched errors
- Created notification store to support global notifications
- Notifications support static messages (in this case, after successfully creating, updating or deleting item of each page, notification is woken up and assigned message corresponding to previous action).
- Fixed bug on test page where not existing function is triggered

PR Resolves: #338, #340 and #341


web/src/main/frontend/src/base/stores/FlareStore.js
```diff
@@ -1,70 +0,0 @@
-import {action, observable} from 'mobx';
-import {TYPE} from '../../components/ui/MsgFlare.js';
-import * as R from 'ramda';
```
iscekic: Let's remove `ramda` as a dependency - we aren't using it enough. 😄

web/src/main/frontend/src/base/stores/QuestionStore.js
```diff
@@ -54,7 +59,20 @@ class QuestionStore {
       ? questionApi.updateQuestion
       : questionApi.createNewQuestion;
 
-    return submit(question).then(() => this.fetchQuestions());
+    return submit(question).then(() => {
+      this.fetchQuestions();
+      setNotificationMessage(
+        question.id
```
iscekic: No need for 2 objects if the only difference is the message.

web/src/main/frontend/src/base/errorHandler.js
```diff
@@ -1,27 +1,15 @@
...
- * By default, all errors derived from API calls (marked with 'apiCallError') are handled
- * as flare error notifications.
- * Other error objects are re-thrown.
- *
- * @param err Error object
- */
-export const handleApiError = err => {
-  if (err.apiCallError) {
-    FlareStore.error(err);
+export const setNotificationMessage = error => {
```
iscekic: `error` is a misleading name, how about `notification`?

vrackovic: I agree. Will be changed to `notification` then.

web/src/main/frontend/src/base/stores/FlareStore.js
vrackovic: Actually, we use its `keys`, `head`, `compose` and `replace` functions in some places. Should it be refactored, or should `ramda` be used in those places?

PR #376, opened 2019-06-10
Fix getAllReviewers

dujfalusi:


core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -230,6 +235,11 @@ public void assignReviewerToTest(Long testId, Long reviewerId) {
     Test test = findOne(testId);
     User user = userRepository.getOne(reviewerId);
 
+    if (!user.isReviewer()){
```
stanislavgrujic: Please configure your formatter.

PR #382, opened 2019-06-25
Changed authenticationUtil so it is a service now

Fistulator:


web/src/main/java/eu/execom/testgenerator/web/service/AuthenticationService.java
```diff
@@ -0,0 +1,22 @@
+package eu.execom.testgenerator.web.service;
+
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken;
+import eu.execom.testgenerator.core.model.User;
+import java.security.Principal;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.core.Authentication;
+
+public interface AuthenticationService {
```
stanislavgrujic: There is no need to introduce an interface for this service. You should have just renamed the `AuthenticationUtil` to `AuthenticationService`. Please remove this interface, and just do the renaming.

web/src/main/java/eu/execom/testgenerator/web/controller/LoginController.java
```diff
@@ -69,6 +68,8 @@ private ResponseEntity handleGoogleLogin(GoogleIdToken idToken) {
     }
 
     User user = userService.getUserByNameAndEmail(name, email, UserRole.EMPLOYEE);
+    System.out.println(user);
+    System.out.println(userService.findByEmail(email));
```
stanislavgrujic: Please use proper logging, if needed, instead of printing to `System.out`. This looks like a debugging leftover, please remove it.

Fistulator: I apologise, leftovers have now been removed.

PR #385, opened 2019-06-26
Fixed issue with test timer

Fistulator:
Users of the Test API can no longer send a Test's remaining time to it. It is now calculated on the server and cannot be modified.


core/src/main/java/eu/execom/testgenerator/core/model/Test.java
```diff
@@ -68,4 +71,8 @@
   @Where(clause = "deleted = 'false'")
   @OneToMany(mappedBy = "test", cascade = CascadeType.ALL)
   private List<TestQuestion> testQuestions;
+
+  public int calculateRemainingTime() {
```
stanislavgrujic: Even though I like DDD approach, to have business domain logic in model classes, it is not consistent with the rest of the code base. Please move this method to a `TimeServiceImpl`. Other option is to create a constructor which will set this value on instance creation.

Fistulator: Fixed.

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -343,4 +342,12 @@ private TestQuestionAnswer createTestQuestionAnswerWithInput(TestQuestion testQu
     testQuestionAnswer.setInput(answerText);
     return testQuestionAnswer;
   }
+
+  public void calculateAndSetRemainingTime(Test test) {
```
stanislavgrujic: Please always use `@Override` annotation for overridden methods.

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -343,4 +342,12 @@ private TestQuestionAnswer createTestQuestionAnswerWithInput(TestQuestion testQu
...
+    test.setRemainingTime(calculateRemainingTime(test.getTestTime(), test.getFormula().getTestDuration()));
+  }
+
+  public int calculateRemainingTime(long startTime, int duration) {
```
stanislavgrujic: Please make this method private. It is not used outside of the class scope.

core/src/main/java/eu/execom/testgenerator/core/service/TestServiceImpl.java
```diff
@@ -343,4 +342,12 @@ private TestQuestionAnswer createTestQuestionAnswerWithInput(TestQuestion testQu
     testQuestionAnswer.setInput(answerText);
     return testQuestionAnswer;
   }
+
+  public void calculateAndSetRemainingTime(Test test) {
+    test.setRemainingTime(calculateRemainingTime(test.getTestTime(), test.getFormula().getTestDuration()));
```
stanislavgrujic: Please do not use chained method calls, it makes reading code more difficult. Extract the inner method call to a variable, which you will pass to `setRemainingTime()`.

Fistulator: Fixed these issues.

Fistulator: JWT dependency was added because I have started working on switching the app to JWT instead of its current tokenstore.

stanislavgrujic: Please move that dependency to a separate branch. It should not be part of this PR.

Fistulator: Dependency moved in newest commit.

web/pom.xml
```diff
@@ -65,6 +65,9 @@
       <version>${google.apis.plus.version}</version>
     </dependency>
 
+
+
+
```
stanislavgrujic: Please remove these empty lines

Issue #360, opened 2018-12-10
Implement answers on new test after reload


vrackovic:
Flow:
- Guest user signs in, new test is generated, and displayed so that user can answer questions
- If user tries to reload page, same test will be loaded. Same applies for back button, and changing url by hand. No matter what, user will be redirected to generated test until test is submitted.
- Every time user answers a question (`onClick` for checkboxes, and `onBlur` for input field questions), answer is sant to BE with PUT method.

Problem: 
- Reloading page gets us test and questions fetched and displayed, but answers are not displayed.

Task:
- After test is reloaded, and test with questions is fetched, fetch answers for that `testId` for that user, and display saved questions.

Notes: 
- `/test/{testId}` doesn't return answers object. It should be fixed on backend so that previously answered questions are returned together with answers, so that answers can be mapped and rendered properly.

dujfalusi:
Backend is updated with this pull request #379.

Fistulator:
Solved in pull request #389

PR #389, opened 2019-07-02
Remember test answers

Fistulator:
Solving issue #360


web/src/main/frontend/src/components/TimeRemaining.js
```diff
@@ -5,6 +5,7 @@ import {inject, observer} from 'mobx-react';
 @observer
 class TimeRemaining extends React.Component {
   componentWillUpdate() {
+    // TODO write onwarning method
```
stanislavgrujic: Do not leave hanging TODOs in code. Either create a task for it and reference it in the code, or clean it up right away.

web/src/main/java/eu/execom/testgenerator/web/dto/TestDto.java
```diff
@@ -33,10 +33,14 @@ public TestDto(Test test) {
...
       this.questions = test.getTestQuestions()
                            .stream()
-                           .map(TestQuestion::getQuestion)
-                           .map(QuestionDto::new)
+                           .map(this::makeQuestionWithTestAnswers)
                            .collect(Collectors.toList());
     }
   }
 
+  private QuestionDto makeQuestionWithTestAnswers(TestQuestion testQuestion) {
```
stanislavgrujic: More expressive would be to name the method `toQuestionWithTestAnswers` or `mapToQuestionWithTestAnswers`.

stanislavgrujic:
@iscekic Please review frontend.

web/src/main/frontend/src/components/TakeTestQuestionComponent.js
```diff
@@ -10,7 +10,21 @@ import {inject, observer} from 'mobx-react';
 class TakeTestQuestionComponent extends React.Component {
   //TODO: review iscekic, These two lines look like they belong in a lifecycle method.
   answers = this.props.newTestForm.$('answers').add();
+
+  prepareMultipleChoiceQuestion() {
+    const checkedAnswers = [...this.props.testAnswers.testQuestionAnswerIds];
```
iscekic: There is no need to create a copy as we aren't mutating anything.

web/src/main/frontend/src/components/TakeTestQuestionComponent.js
```diff
@@ -10,7 +10,21 @@ import {inject, observer} from 'mobx-react';
...
+    const allAnswerIds = [...this.props.question.answers].map(obj => obj.id);
```
iscekic:
```suggestion
    const allAnswerIds = this.props.question.answers.map(obj => obj.id);
```

`map` returns a new array without touching the mapped one.

web/src/main/frontend/src/components/TakeTestQuestionComponent.js
```diff
@@ -10,7 +10,21 @@ import {inject, observer} from 'mobx-react';
...
+    const andswersAndStatus = allAnswerIds.reduce((accumulator, id) => {
+      const isInCheckedAnswers = checkedAnswers.indexOf(id) !== -1;
+      accumulator[id] = isInCheckedAnswers;
```
iscekic: You can inline this variable.

web/src/main/frontend/src/components/ui/Checkbox.js
```diff
@@ -15,6 +15,8 @@ const Checkbox = props => {
     }
   }
 
+  propsContainer.defaultChecked = props.defaultChecked;
```
iscekic: I dislike this entire component. Can you refactor it to be more in line with `mobx-react-form`:

```js
// example usage
<Checkbox {...form.$('whatever').bind()} someProp="1" />
```

web/src/main/frontend/src/components/TakeTestQuestionComponent.js
```diff
@@ -10,7 +10,21 @@ import {inject, observer} from 'mobx-react';
 class TakeTestQuestionComponent extends React.Component {
   //TODO: review iscekic, These two lines look like they belong in a lifecycle method.
```
iscekic: Take a look at this `TODO` and see if you can fix it as part of this task.

web/src/main/frontend/src/components/TakeMultipleChoiceQuestion.js
```diff
@@ -7,18 +7,14 @@ import Checkbox from './ui/Checkbox';
...
-      [option.id]: this.props.isSelected
-    });
     return (
       <Checkbox
         key={option.id}
         label={option.description}
         formField={answer.bind({
           value: option.id
         })}
+        defaultChecked={answer.values()[option.id]}
```
iscekic: See comment above. 💃

web/src/main/frontend/src/components/TakeTestQuestionComponent.js
Fistulator: I needed an array of checked answers, using .values() on testQuestionAnswerIds gives me an iterator, using it without values() gives me a much larger object than needed.

web/src/main/frontend/src/components/ui/Checkbox.js
Fistulator:
```suggestion
  <Checkbox
        answer = {answer}
        key = {option.id}
        option={option}
      />
```
Would something like this be better? Checkbox's code would look like:

```suggestion
<div className="checkbox_wrapper">
      <label>
        <input {...props.answer.bind(
          {
            value: props.option.id
            })} 
            defaultChecked={checked} 
            type="checkbox" />
        <div className="label_icon">
          <span className={answerTextClassName}>{props.option.description}</span>
        </div>
      </label>
    </div>
```

Fistulator: Not sure I understand what I need to do.

web/src/main/frontend/src/components/TakeTestQuestionComponent.js
iscekic: My comment was about not copying `testQuestionAnswerIds` into a new variable `checkedAnswers` as we aren't changing it anywhere. No change == no need to copy.

web/src/main/frontend/src/components/ui/Checkbox.js
iscekic: I'd like the checkbox component to behave as much as eg. a simple input would:

```js
// https://foxhound87.github.io/mobx-react-form/docs/quick-start.html
<input {...form.$('whatever').bind()} />
```

web/src/main/frontend/src/components/TakeTestQuestionComponent.js
Fistulator: Understood.

web/src/main/frontend/src/components/ui/Checkbox.js
Fistulator: Upon looking further at the code, I realized that CheckBox was used by two different forms and that their functionality was relatively different. I decided to remove the component entirely as it was not doing much already and making it work with different form types would require either adding an addiotional component type that would be used only once or making some ugly code to make it comply with everything.

web/src/main/frontend/src/components/Test.js
```diff
@@ -55,6 +55,8 @@ class Test extends React.Component {
   //TODO vrackovic: Investigate implementation and fill this function if needed
   handleAlarm = () => {};
 
+  handleWarning = () => {};
```
iscekic: Should this be empty without a `// TODO`?

Fistulator: Will write an issue for these functions.

web/src/main/java/eu/execom/testgenerator/web/dto/TestQuestionAnswerDto.java
```diff
@@ -16,7 +15,7 @@
   private String input;
   private List<Long> testQuestionAnswerIds;
 
-  TestQuestionAnswerDto(TestQuestion testQuestion) {
+  public TestQuestionAnswerDto(TestQuestion testQuestion) {
```
stanislavgrujic: Why did you change the access modifier? It is not used outside the dto package, as far as I see it?

Fistulator: I believe it caused an error, but the access modifier could not have been the cause (as everything works as it should after removing the access modifier)

Restored it to its original value now.

Fistulator:
@iscekic has already reviewed the frontend part.

Issue #403, opened 2020-04-06
Add default user


stanislavgrujic:
Add database migration to add the default user "hr@execom.eu".

PR #404, opened 2020-04-06
Add default user(s) #403

draganakovacevic:
Add database migration for default user(s) using XML and Change pom.xml file with XML dependency.


core/pom.xml
```diff
@@ -12,6 +12,14 @@
   <artifactId>test-generator-core</artifactId>
 
   <dependencies>
+
+    <!-- XML Db migration -->
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+      <version>2.3.0</version>
+    </dependency>
```
stanislavgrujic, line 21: This was not needed until now, and you are not introducing anything new, conceptually., JAXB is used mostly for serialization/deserialization of objects to XML. Why do you need it?

core/src/main/resources/db/changelog/db.changelog-initializeDatabaseSchema.xml
```diff
@@ -326,4 +326,18 @@
     </createTable>
 
   </changeSet>
+  <changeSet author="application" id="database_initialization2">
```
stanislavgrujic, line 329: Id is misleading. If we have our application in production for 2 years, and have frequent updates to the database, we will come to `database_initialization40` and will not know what each of those migrations is. Please try to be more specific with naming to make it easier to read code and track back in history. I would name this id as `add_default_user`.

core/src/main/resources/db/changelog/db.changelog-initializeDatabaseSchema.xml
```diff
@@ -326,4 +326,18 @@
...
+      <column name="id" value="42"/>
+      <column name="name" value="Dunja Vuic"/>
+      <column name="deleted" valueBoolean="false"/>
+      <column name="email" value="hr@execom.eu"/>
+    </insert>
+    <insert tableName="user">
+      <column name="id" value="43"/>
+      <column name="name" value="Ivana Drakulic"/>
+      <column name="deleted" valueBoolean="false"/>
+      <column name="email" value="hr@execom.eu"/>
```
stanislavgrujic, line 340: Users should have unique email, there should not be two users with the same email. If the database allows it, please also add a migration which adds a unique constraint on `email` column in the `user` table. `hr@execom.eu` is a generic email, which is inherited/shared between persons working in our HR department.

core/src/main/resources/db/changelog/db.changelog-initializeDatabaseSchema.xml
```diff
@@ -326,4 +326,18 @@
     </createTable>
```
stanislavgrujic, line 326: The idea of migrations is that it allows you to change the database schema without dropping everything. For a system which is already running, and which has already run this changelog, it will never be executed again, unless you drop the database completely, which will cause the loss of data. Please add a new migration script for this change.
Read these articles, if you haven't already
[Schema migration - Wikipedia](https://en.wikipedia.org/wiki/Schema_migration)
[How Liquibase works](https://www.liquibase.org/get_started/how-lb-works.html)
[Why database schema migrations - Flyway](https://flywaydb.org/getstarted/why)

core/src/main/resources/db/changelog/db.changelog-initializeDatabaseSchema.xml
```diff
@@ -326,4 +326,18 @@
...
 
   </changeSet>
+  <changeSet author="application" id="database_initialization2">
+    <insert tableName="user">
+      <column name="id" value="42"/>
+      <column name="name" value="Dunja Vuic"/>
```
stanislavgrujic, line 332: Please use HR, since this email is used for all HR employees.

core/src/main/resources/db/changelog/db.changelog-initializeDatabaseSchema.xml
```diff
@@ -326,4 +326,18 @@
...
+      <column name="deleted" valueBoolean="false"/>
+      <column name="email" value="hr@execom.eu"/>
```
stanislavgrujic, line 334: Did you try running this migration? Did you check if users are added to the database? If not, please always test your changes before you create a pull request and mark your task as done. It seems to me that this migration cannot even pass because you are missing `role` column value which is a non-null column. Please add role ADMINISTRATOR.

core/pom.xml
draganakovacevic: Without this dependency, I got exception for entityManager.

Issue #400, opened 2020-04-06
Copy-paste command issue


VojislavKovacevic:
Copy-paste command (ctrl+v) is disabled for Execom users in a number of fields on "Category", "Question" and "Formula" screens. This happens if the text is copied from different fields. The message "No cheating! :)" popup. 
If text is copied from outside of the Test generator, the paste is allowed.
On the test execution screen (non-Execom user), it is possible to use copy-paste commands.

VojislavKovacevic:
Copy-paste command (ctrl+v) is still disabled for Execom users in a number of fields on "Category", "Question" and "Formula" screens. This happens if the text is copied from different fields. The message "No cheating! :)" popup.
If text is copied from outside of the Test generator, the paste command is allowed.
There should not be any restriction for Execom employees in regard to copy/paste functions.

On the test execution screen (non-Execom user), it is possible to use copy-paste commands if the text is copied from a source outside of the test generator app.
The applicant should be forbidden to use copy-paste commands.

PR #410, opened 2020-04-15
Copy paste is allowed for Execom employees #400

draganakovacevic:
Copy paste is still disabled for test takers, but for Execom employees is allowed.


stanislavgrujic:

web/src/main/frontend/src/index.js
```diff
@@ -24,8 +24,10 @@ import {inviteEmployeeForm} from './components/forms/InviteEmployeeForm';
 
 class Root extends React.Component {
   handleCopyEvent(e) {
-    e.clipboardData.setData('text/plain', 'No cheating! :)');
-    e.preventDefault();
+    if(GuestStore.testId != null && GuestStore.testId != '') {
```
BraneOpacic21, line 27: Not sure what data structures are available for `GuestStore.testId`, but if there are 2 options: `null` and a string value (empty or populated) - then this should probably be enough:

`if(GuestStore.testId) {
// stuff 
}`

or

`if(!GuestStore.testId) {
// stuff 
}`

Depending on the logic you are trying to implement, right?

web/src/main/frontend/src/scss/components/_formula.scss
```diff
@@ -1,4 +1,7 @@
 .formula__block {
+
```
BraneOpacic21, line 2: Remove empty rows like these in your CSS/SCSS files - hint: use an extension like prettier or beautify to aid you

web/src/main/frontend/src/scss/components/_formula.scss
```diff
@@ -48,12 +51,13 @@
...
 
   .formula__categories_block {
-    max-height: 240px;
+    position: inherit;
     margin-top: 20px;
-    overflow-y: scroll;
     margin-bottom: 20px;
     width: 100%;
-    z-index: 10;
+    z-index: 15;
```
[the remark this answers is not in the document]
BraneOpacic21: Further explanation: Skipping numbers would potentially lead to a bigger problem once the project scales, leaving you frustrated trying to 'debug' CSS

web/src/main/frontend/src/scss/components/_formula.scss
```diff
@@ -48,12 +51,13 @@
...
+    display: inline-block;
+    flex: 1;
```
BraneOpacic21, line 60: /n

Separate your CSS rules, please

Issue #399, opened 2020-04-06
In the field "Points" on the "Question" screen, it is possible to enter negative numbers and special characters.


VojislavKovacevic:
In the field "Points" on the "Question" screen, it is possible to enter negative numbers and special characters.

The field should be limited to accept positive numbers only.

PO: Are decimal values requiered? 

![formula add recipe2020-04-02 12_04_25-Execom test](https://user-images.githubusercontent.com/63228397/78540387-621cc680-77f4-11ea-8afb-bd3ac08de26b.png)

draganakovacevic:
Special characters for points in Question should not be possible to enter because field was numeric type. Screenshot for this bug is not correct because this is connected with Question not Formula form.

PR #411, opened 2020-04-16
Fix value of points in Question #399

draganakovacevic:
There was option to enter decimal numbers but now it's fix to only accept integer numbers. Range of numbers cannot be negative because it was previously set with @Min and @Max annotations.


stanislavgrujic:

web/src/main/frontend/src/scss/layout/_panel.scss
```diff
@@ -213,12 +218,16 @@ $border: 1px solid white;
...
+  z-index: 1;
+  position: fixed;
+  bottom: 180px;
   width: 100%;
+  height: 75px;
   button.icon__block {
     margin-left: -15px;
+    position: inherit;
+    height: initial;
+    z-index: inherit;
```
BraneOpacic21, line 230: Please, double-check if these 3 lines are needed at all

web/src/main/frontend/src/scss/components/_formula.scss
```diff
@@ -48,12 +51,13 @@
...
 
   .formula__categories_block {
-    max-height: 240px;
+    position: inherit;
     margin-top: 20px;
-    overflow-y: scroll;
     margin-bottom: 20px;
     width: 100%;
-    z-index: 10;
+    z-index: 15;
```
BraneOpacic21, line 58: As we discussed prior, lets increment the `z-index` by one each time, so we don't end up (as our app grows) with unnecessary large numbers which are hard to track

web/src/main/frontend/src/index.js
```diff
@@ -24,8 +24,10 @@ import {inviteEmployeeForm} from './components/forms/InviteEmployeeForm';
 
 class Root extends React.Component {
   handleCopyEvent(e) {
-    e.clipboardData.setData('text/plain', 'No cheating! :)');
-    e.preventDefault();
+    if(GuestStore.testId != null && GuestStore.testId != '') {
+      e.clipboardData.setData('text/plain', 'No cheating! :)');
+      e.preventDefault();
+    }
```
BraneOpacic21, line 30: I suppose that we would want to _prevent the default behavior_ of the event regardless of the `if statement`, right? 🤔  Correct me if I'm wrong.

Also, the condition could probably be written a bit simpler:

`if(GuestStore.testId) {
// do something
}`

or

`if(!GuestStore.testId) {
// do something
}`

depending on the implementation you choose

web/src/main/frontend/src/scss/layout/_panel.scss
```diff
@@ -213,12 +218,16 @@ $border: 1px solid white;
 }
 
 .panel__add_button {
-  z-index: 8;
-  position: absolute;
-  bottom: 60px;
+  z-index: 1;
```
BraneOpacic21, line 221: Double-check please, if `z-index` is not necessary here, remove it

Issue #398, opened 2020-04-06
In the field "Duration" on "Formula screen", it is possible to enter negative numbers


VojislavKovacevic:
In the field "Duration" on "Formula screen", it is possible to enter negative numbers and 0. This will cause the instant closing of the test screen for the participant. 
The minimal duration of the test should be fixed (30 minutes? need confirmation from Product owner)

![formula add recipe2020-04-02 12_04_25-Execom test](https://user-images.githubusercontent.com/63228397/78540326-4c0f0600-77f4-11ea-9822-4a18d667b528.png)

PR #412, opened 2020-04-16
Fix value of duration in Formula #398

draganakovacevic:
Duration in Formula accepted decimal numbers, but now is fixed with integer numbers and also is not possible to enter a negative number because @Min and @Max annotation are set.


stanislavgrujic:

PR #430, opened 2020-05-18
Move mail properties out of properties and Change formatter #427

draganakovacevic:


conf/formatter.xml
```diff
@@ -26,14 +31,6 @@
...
     <option name="ARRAY_INITIALIZER_WRAP" value="1" />
-    <option name="VARIABLE_ANNOTATION_WRAP" value="2" />
-    <indentOptions>
-      <option name="INDENT_SIZE" value="2" />
-      <option name="CONTINUATION_INDENT_SIZE" value="4" />
-      <option name="TAB_SIZE" value="2" />
-    </indentOptions>
-  </codeStyleSettings>
-  <codeStyleSettings language="XML">
     <indentOptions>
```
stanislavgrujic, line 34: Why is this file changed?

core/src/main/resources/application.properties
```diff
@@ -24,18 +24,5 @@ security.oauth2.client.scope=profile email
 security.oauth2.resource.userInfoUri=https://www.googleapis.com/userinfo/v2/me
 security.oauth2.resource.preferTokenInfo=false
 
-spring.mail.host=smtp.gmail.com
-spring.mail.port=587
-spring.mail.username=test.gen.app@gmail.com
-spring.mail.password=fqlkyqviswipbskv
-spring.mail.properties.mail.smtp.auth=true
-spring.mail.properties.mail.smtp.starttls.enable=true
```
stanislavgrujic, line 32: Spring configuration should not be taken out of application.properties file.

core/src/main/resources/application.yml
```diff
@@ -0,0 +1,21 @@
...
+   username: test.gen.app@gmail.com
+   password: fqlkyqviswipbskv
+   properties:
+     mail:
+       smtp:
+         auth: true
+         startls:
+           enable: true
+
+email:
```
stanislavgrujic, line 14: This should go into email template files, not into properties. Using yml files is just the same as using properties files, only in different format. What we want is to have templates for each of emails. Then we can beautify those emails above plain text. Please read these two articles and try to understand the difference between configuration and templates.
[Guide to Spring Email](https://www.baeldung.com/spring-email)
[Using ThymeLeaf and FreeMarker Emails Templates with Spring](https://www.baeldung.com/spring-email-templates)

conf/formatter.xml
draganakovacevic: I  just set format which we got in mail on this file.

draganakovacevic:
I returned spring properties in application file, and also still have file with email's information. These information I use in Email service to work with and add dependency for FreeMarker. In resource there are 2 files for FreeMarker's templates.

web/src/main/java/eu/execom/testgenerator/web/controller/LoginController.java
```diff
@@ -61,9 +61,9 @@ private ResponseEntity handleGoogleLogin(GoogleIdToken idToken) {
     String name = (String) payload.get("name");
     String domain = (String) payload.get("hd");
 
-    if (!EXECOM_DOMAIN.equals(domain)) {
-      throw new AccessDeniedException("Email not from Execom domain!");
-    }
+    //if (!EXECOM_DOMAIN.equals(domain)) {
+      //throw new AccessDeniedException("Email not from Execom domain!");
+    //}
```
stanislavgrujic, line 66: Please do not leave commented out code. I assume this was commented out only for testing purposes. Should it be reverted?

stanislavgrujic:
Please resolve merge conflicts.

Issue #432, opened 2020-05-19
Replace @Data and @NoArgsConstructor with @Value in classes 


draganakovacevic:

VojislavKovacevic:
@draganakovacevic @stanislavgrujic 

Is this issue testable from a user perspective?
Is this issue merged to a develop branch?

PR #433, opened 2020-05-19
Replace @Data and @NoArgsConstructor with @Value #432 

draganakovacevic:


web/src/main/java/eu/execom/testgenerator/web/dto/GetAllCategoriesResponse.java
```diff
@@ -1,17 +1,11 @@
...
+import lombok.Value;
 
 import java.util.List;
 
-@NoArgsConstructor
-@Data
+@Value
 public class GetAllCategoriesResponse {
 
     private List<CategoryResponse> categoryResponses;
```
stanislavgrujic, line 10: Please also remove all `private` properties since `@Value` generates it.

stanislavgrujic:
Please update with the latest develop

Issue #441, opened 2020-05-26
It is possible to assign a negative number to the number of questions in formula


VojislavKovacevic:
It is possible to assign a negative number to the number of questions in formula.

Example image: The number of questions for category C1 is -2. The formula is successfully saved.

![ReceipeNegativNumber2020-05-26 11_10_27-Execom test](https://user-images.githubusercontent.com/63228397/82885191-17dada80-9f45-11ea-9715-bb3184e14444.png)

PR #443, opened 2020-05-26
Fix number of questions in Formula #441

draganakovacevic:


core/src/main/java/eu/execom/testgenerator/core/service/FormulaServiceImpl.java
```diff
@@ -42,8 +42,12 @@ public Formula save(Formula formula) {
     }
 
     formula.getRecipes().forEach(recipe -> {
-      recipe.setId(null);
-      recipe.setFormula(formula);
+      if(recipe.getNumberOfQuestions() < 0) {
+        throw new ValidationException("Number of questions in recipe can not be less or equals 0.");
```
stanislavgrujic, line 46: Can't you just add the validation to the RecipeDto to use `@Min(value = 0)` along with `@NotNull`?

PR #446, opened 2020-05-27
Sequence update fix

adamovivan:


core/src/main/resources/db/changelog/db.changelog-2.1.xml
```diff
@@ -0,0 +1,17 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="sequence_update" author="iadamov">
+        <comment>Updates user and granted_user_role sequences</comment>
+
+        <insert tableName="hibernate_sequences">
+            <column name="sequence_name" value="user"/>
+            <column name="sequence_next_hi_value" value="1"/>
```
stanislavgrujic, line 10: Didn't we agree that you will just omit ids in `changelog-2.0`? Does that not work?

adamovivan: Yes, but it doesn't work. Ids must be specified. There are some workarounds but this solution is better.

PR #453, opened 2020-05-29
Error "no suitable constructor found"

adamovivan:


web/src/main/java/eu/execom/testgenerator/web/dto/AcceptInviteDto.java
```diff
@@ -1,15 +1,17 @@
...
 
-import lombok.Value;
+import lombok.Getter;
+import lombok.Setter;
 
 import javax.validation.constraints.NotNull;
 
-@Value
+@Getter
+@Setter
```
stanislavgrujic, line 9: Is there a reason not to use `@Data` annotation?

Issue #417, opened 2020-04-22
proposal: add field for reviewer comments


VojislavKovacevic:
proposal: add field for reviewer comments

Field for reviewer comments should be added under every answer in the same line with points field.
Field for general comment should be added at the end of the test in the same line with the result field.

Comment field should be a string type, same format as questions. 
Comment fields should be saved when a review is submitted.

VojislavKovacevic:
@stanislavgrujic please review the proposal and confirm if it is according to requirements.

PR #456, opened 2020-06-02
 proposal: add field for reviewer comments #417 

adamovivan:


web/src/main/frontend/src/base/stores/QuestionStore.js
```diff
@@ -7,17 +7,14 @@ class QuestionStore {
...
   @observable answers = [];
-
   @observable selectedCategory = '';
 
   @action.bound
   setActiveQuestionById(id) {
-    this.questions.find(
-      question => (question.id === id ? (this.activeQuestion = question) : null)
+    this.questions.find(question =>
+      question.id === id ? (this.activeQuestion = question) : null
```
BraneOpacic21, line 17: This could probably be `this.activeQuestion = this.questions.find(question => question.id === id)` if you could set `undefined` instead of `null` if there are not id matches here, right? 

Not a mistake, just increasing readability here.

Double-check if we could use the more readable solution, please.

web/src/main/frontend/src/scss/components/_reviewTest.scss
```diff
@@ -157,3 +157,23 @@ textarea {
...
+}
+
+.result-comment {
+  margin-bottom: 30px;
+  text-align: right;
+  textarea {
+    width: 40%;
+    height: 75px;
+    margin-top: 10px;
+  }
```
BraneOpacic21, line 178: A lot of CSS is same here, we could use the **base CSS class** for both, and then only add the **different parts afterwards**

core/src/main/java/eu/execom/testgenerator/core/service/TestService.java
```diff
@@ -31,7 +30,8 @@
 
   void remove(Long id);
 
-  void applyReview(Test test, User user, Map<Long, Integer> mapQuestionIdToAwardedPoints);
+  void applyReview(Test test, User user, Map<Long, Integer> mapQuestionIdToAwardedPoints,
```
stanislavgrujic, line 33: Instead of having 3 different parameters, I suggest to create a structure which will cover all of them:
```
@Data
public class TestReview {
  private User reviewer;
  private String comment;
  private Map<Long, QuestionReview> questionsReviews;

  public QuestionReview getQuestionReview(long questionId) {
    return questionReviews.get(questionId);
  }

}
@Value
private QuestionReview {
  long questionId;
  int points;
  String comment;
}

// then you can have method signature
void applyReview(Test test, TestReview review) {
  test.getTestQuestions().forEach(question -> updateWithReview(question, review.getQuestionReview(question.getId());
}
```

web/src/main/java/eu/execom/testgenerator/web/dto/CreateQuestionRequest.java
```diff
@@ -32,5 +32,7 @@
 
   private String codeImage;
 
+  private String comment;
```
stanislavgrujic, line 35: Should this be here? We are not adding comments to questions when we create them.

web/src/main/java/eu/execom/testgenerator/web/dto/SubmitReviewRequest.java
```diff
@@ -15,6 +15,7 @@
    */
   @AllAnswersReviewed
   private Map<Long, Integer> scores;
-
   private Long testId;
+  private Map<Long, String> questionComments;
```
stanislavgrujic, line 19: Please place this field below scores, since they are similar in usage. Also, name them in similar way (`scores` & `comments`).

web/src/main/java/eu/execom/testgenerator/web/dto/SubmitTestRequest.java
```diff
@@ -11,4 +11,5 @@
 
   private Long testId;
   private Map<Long, TestQuestionAnswerDto> answers;
+  private String comment;
```
stanislavgrujic, line 14: Why did you add this field? It is not used anywhere. Please remove it.

web/src/main/java/eu/execom/testgenerator/web/dto/UpdateQuestionRequest.java
```diff
@@ -29,5 +29,7 @@
 
   private String codeImage;
 
+  private String comment;
```
stanislavgrujic, line 32: Why did you add this field? It is not used anywhere. Please remove it.

web/src/main/java/eu/execom/testgenerator/web/util/Constants.java
```diff
@@ -7,5 +7,5 @@ private Constants() { }
   public static final String AUTHENTICATION_TOKEN_KEY = "X-AUTH-TOKEN";
   public static final String APPLICANT_TOKEN_KEY = "X-GUEST-TOKEN";
   public static final String EXECOM_DOMAIN = "execom.eu";
-  public static final String USER_IS_ADMIN = "USER_IS_ADMIN";
+  public static final String USER_IS_ADMIN = "USER-IS-ADMIN";
```
stanislavgrujic, line 10: In JavaScript code this is lowercase. Is matching strings in JS case-insensitive?

adamovivan: This constant will be removed, it's unnecessary because roles are added in jwt token.

core/src/main/java/eu/execom/testgenerator/core/service/granted_user_role/GrantedUserRoleService.java
```diff
@@ -1,4 +1,4 @@
-package eu.execom.testgenerator.core.service;
+package eu.execom.testgenerator.core.service.granted_user_role;
```
stanislavgrujic, line 1: Are you sure this requires a separata package? When I think about packages, I think "Can this entity exists as a standalone entity?". I don't think that this one can, so I would place it either in the `user` package, or the `role` package.

core/src/main/java/eu/execom/testgenerator/core/service/question/QuestionServiceImpl.java
```diff
@@ -15,7 +16,7 @@
 import java.util.stream.Collectors;
 
 @Service
-class QuestionServiceImpl implements QuestionService {
+public class QuestionServiceImpl implements QuestionService {
```
stanislavgrujic, line 19: This should not be public. Interface is public, implementation is package private.

core/src/main/java/eu/execom/testgenerator/core/service/category/CategoryServiceImpl.java
```diff
@@ -11,7 +12,7 @@
 import java.util.List;
 
 @Service
-class CategoryServiceImpl implements CategoryService {
+public class CategoryServiceImpl implements CategoryService {
```
stanislavgrujic, line 15: Why this change? Interface should be public, the impelemtation should be package private.

core/src/main/java/eu/execom/testgenerator/core/service/granted_user_role/GrantedUserRoleServiceImpl.java
```diff
@@ -1,4 +1,4 @@
-package eu.execom.testgenerator.core.service;
+package eu.execom.testgenerator.core.service.granted_user_role;
```
stanislavgrujic, line 1: Please be consistent with the current naming convention and in packages do not use underscore to separate words, but rather make it all together, such as `testgenerator` is.

core/src/main/java/eu/execom/testgenerator/core/service/test/QuestionReview.java
```diff
@@ -0,0 +1,17 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+public class QuestionReview {
+  private Long questionId;
+  private Integer points;
+  private String comment;
+
+  public QuestionReview(Long questionId, Integer points) {
```
stanislavgrujic, line 13: You can make questionsId and points `@NonNull`, add `@RequiredArgsConstructor` and it should generate this one.

core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -41,7 +45,7 @@
 
 @Slf4j
 @Service
-class TestServiceImpl implements TestService {
+public class TestServiceImpl implements TestService {
```
stanislavgrujic, line 48: Why this change? Interface should be public, the implementation should be package private.

core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -398,4 +400,25 @@ public void calculateAndSetRemainingTime(Test test) {
   private int calculateRemainingTime(long startTime, int duration) {
     return Math.toIntExact(startTime + duration * 60 - LocalDateTime.now().toEpochSecond(ZoneOffset.UTC));
   }
+
+  private void updateWithReview(TestQuestion question, QuestionReview review) {
+    question.setAchievedPoints(review.getPoints());
+    question.setComment(review.getComment());
+  }
+
+  private void canTestBeReviewedCheck(Test test, User reviewer) {
```
stanislavgrujic, line 409: Please rename to `validateTestCanBeReviewed`. Pay attention to meaning of the method name - if name contains `can`, `has`, `is` or any other wording that would indicate yes/no answer, assumption is that the method returns boolean. Since it is not the case, common verb to use for naming is *validate*.

core/src/main/java/eu/execom/testgenerator/core/service/user_invite/UserInviteService.java
```diff
@@ -1,4 +1,4 @@
-package eu.execom.testgenerator.core.service;
+package eu.execom.testgenerator.core.service.user_invite;
```
stanislavgrujic, line 1: This can probably go to `user` package.

web/src/main/java/eu/execom/testgenerator/web/util/Constants.java
stanislavgrujic: I am well aware of it. For the sake of paying attention to details and making sure the code does not break, one needs to make sure that the introduced changes don't break the code, or at least are not useless. This is a useless change, and as such does not need to exist in the PR/code base, even if it will be removed. If something can break the code, it should not be in the codebase, because you can not rely on it that it will be fixed in the next pull request.

core/src/main/java/eu/execom/testgenerator/core/service/test/QuestionReview.java
adamovivan: Data annotation generates RequiredArgsConstructor

core/src/main/java/eu/execom/testgenerator/core/service/granted_user_role/GrantedUserRoleService.java
adamovivan: Why should I put in the _role_ package when an entity can't exist as a standalone entity? :)

Issue #461, opened 2020-06-03
Aplicant log in 


VojislavKovacevic:
If applicant submit wrong credentials in the login page, an empty popup message is shown.
![popEmpty2020-06-03 16_28_51-Execom test](https://user-images.githubusercontent.com/63228397/83649604-ac81b000-a5b7-11ea-8a95-a29261bd1a07.png)

PR #464, opened 2020-06-04
Applicant login  #461

adamovivan:


web/src/main/frontend/src/base/stores/GuestStore.js
```diff
@@ -19,10 +19,17 @@ class GuestStore {
...
       })
       .catch(err => {
-        setNotificationMessage({
-          type: 'warning',
-          message: err.response.text
-        });
+        if (err.response.status === 404) {
+          setNotificationMessage({
+            type: 'error',
+            message: 'Bad email or code'
```
aberonja, line 25: Maybe use ternary if in the message since setNotificationMessage gets called anyway?

Issue #460, opened 2020-06-03
Formula update error


VojislavKovacevic:
Unexpected behavior during formula update.
![captured](https://user-images.githubusercontent.com/63228397/83644507-a5f03a00-a5b1-11ea-8b94-b9094e8ca6fd.gif)

PR #465, opened 2020-06-04
Formula update error #460

adamovivan:


web/src/main/java/eu/execom/testgenerator/web/dto/UpdateFormulaRequest.java
```diff
@@ -16,10 +20,14 @@
   private Long id;
 
   @NotNull
+  @Size(min = 5, max = 255, message = "Description length should be between 5 and 255 characters.")
```
stanislavgrujic, line 23: Nicer message is -> *Description should be between 5 and 255 characters long."

Issue #457, opened 2020-06-03
Review test design


adamovivan:
- Field for reviewer comments should be added under every answer in the same line with points field.
- Consider the best position for the general comment field.
- _Result_ should not be _readonly_ input type (consider a  bold string instead)

PR #468, opened 2020-06-08
Redesign review test #457

draganakovacevic:


stanislavgrujic:
Please update with the latest develop.

web/src/main/frontend/src/components/ReviewTest.js
```diff
@@ -50,15 +50,16 @@ class ReviewTest extends React.Component {
...
-              <span>RESULT: </span>
-              <input
-                className="result"
-                type="text"
-                value={`${this.props.completedTest.totalScore} / ${this.props.completedTest.perfectScore}`}
-                readOnly
-              />
+              <ReviewTestResultComment />
+              <div className="result-centered-element">
+                <span>RESULT: </span> <br />
```
aberonja, line 55: Do we need to use `<br />`?

Can we do it with a wrapping div?

web/src/main/frontend/src/scss/components/_reviewTest.scss
```diff
@@ -179,3 +179,27 @@ textarea {
...
+
+.review-question-comment {
+  margin-right: 5px;
+  width: 1065px;
+  margin-left: 17px;
+}
+
+.review-question-comment-div {
+  z-index: 5;
+  margin-top: -29px;
```
aberonja, line 200: Why do we have negative margin?

web/src/main/frontend/src/scss/components/_reviewTest.scss
```diff
@@ -179,3 +179,27 @@ textarea {
...
+}
+
+.result-opinion-test {
+  margin-right: 5px;
+  width: 1020px;
+}
+
+.review-question-comment {
+  margin-right: 5px;
+  width: 1065px;
```
aberonja, line 194: This also looks a bit odd. Could we do it with flex? Is this responsive?

web/src/main/frontend/src/scss/components/_reviewTest.scss
draganakovacevic: They are on separate divs because of components, and I want to overlap to be in the same row.

web/src/main/frontend/src/scss/components/_reviewTest.scss
draganakovacevic: I tried with flex and change didn't do wanted "design". This is without

:

stanislavgrujic:
@adamovivan Please resolve merge conflicts and you can merge.

Issue #462, opened 2020-06-03
Submitted tests not visible on the dashboard.


VojislavKovacevic:
Submitted tests not visible on the dashboard.
Applicant is informed that test is submitted.
Test status in DB is STARTED.
Answers are not recorded in DB

PR #472, opened 2020-06-09
Fix view of tests on Dashboard #462

draganakovacevic:
The reason why it was kept as started is because it was not correct path to submit endpoint, where is status set on finished. Also, queries are changed because admin can see all tests even started, etc. and for Reviewer is important to assign him, where we change status of test.


stanislavgrujic:
Please update with the latest develop.

stanislavgrujic:
Please resolve merge conflicts and try not to override valid code.

stanislavgrujic:
@adamovivan Could you please resolve merge conflicts?

PR #489, opened 2020-06-15
Fix Multiple choices questions

draganakovacevic:
Fix warnings for equivalence


core/src/main/java/eu/execom/testgenerator/core/service/question/QuestionServiceImpl.java
```diff
@@ -156,10 +157,14 @@ public void remove(Long id) {
   }
 
   private boolean isValid(Question question) {
+    boolean valid = true;
     if (question.isMultipleChoice()) {
-      return question.getAnswers().stream().anyMatch(answer -> answer.isActive() && answer.isCorrect());
+      valid = question.getAnswers().stream().anyMatch(answer -> answer.isActive() && answer.isCorrect());
+      if(!valid) {
+        throw new ValidationException("Question must have a checked correct answer. ");
+      }
```
stanislavgrujic, line 165: This change will prevent saving invalid questions, while the issue description says *"The system should not include such a question when the validity of the formula is checked."*.
Which one is the correct behaviour?

core/src/main/java/eu/execom/testgenerator/core/service/question/QuestionServiceImpl.java
```diff
@@ -156,10 +157,14 @@ public void remove(Long id) {
...
     }
-    return true;
+    return valid;
```
stanislavgrujic, line 167: Why this change when the method always returns *true* and throws an exception in case that it is *false*?

core/src/main/java/eu/execom/testgenerator/core/service/question/QuestionServiceImpl.java
draganakovacevic: Understand, i set that we need correct answer for question to be valid.

PR #497, opened 2020-06-16
Fix saving just created Test and Set retrieving all Tests without con…

draganakovacevic:
First we create test 

After, when applicant starts to doing created test

And the end, when applicant submit his test we see that test is finished


core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -124,6 +124,8 @@ public Test create(String email, String applicantName, Long formulaId) {
 
     test.setMaximumPoints(maximumPoints);
     test.setAchievedPoints(0);
+    test.setTimeTaken(LocalDateTime.now());
+    testRepository.save(test);
```
stanislavgrujic, line 128: Is it not that the Test is saved in a cascade when User is saved? Is this change really needed?

draganakovacevic: Mistake, yes it's saved cascade with user. TimeTaken for status "CREATED" is when test is created, after that it will be updated with time when test is started.

PR #500, opened 2020-06-18
Feature branch pbi463 second

draganaHrcek:


core/src/main/java/eu/execom/testgenerator/core/service/question/QuestionService.java
```diff
@@ -7,15 +7,18 @@
 
 public interface QuestionService {
 
-  Question save(Question question, Long categoryId);
+	Question save(Question question, Long categoryId);
```
stanislavgrujic, line 10: Please setup your IDE to use Execom formatter and revert back changes in white space.

core/src/test/java/eu/execom/testgenerator/core/service/question/QuestionServiceTest.java
```diff
@@ -43,12 +46,14 @@
   @InjectMocks
   private QuestionServiceImpl questionService;
 
+  private List<Question> allQuestions;
```
stanislavgrujic, line 49: This is dangerous because it can easily stay in a certain state between different tests, which can lead to having successful execution of one test depend on execution of another test. Having a state in between tests, for which you have to take care to leave it in such a way that the following tests can pass will lead to complex code with high cohesion between tests which makes it very difficult for maintenance. Please use local variable instead.

web/src/main/java/eu/execom/testgenerator/web/controller/QuestionController.java
```diff
@@ -41,8 +42,15 @@ public QuestionController(QuestionService questionService) {
...
   @PostMapping
-  public ResponseEntity<CreateQuestionResponse> createNewQuestion(@RequestBody @Valid CreateQuestionRequest createQuestionRequest) {
+  public ResponseEntity<CreateQuestionResponse> createNewQuestion(
+      @RequestBody @Valid CreateQuestionRequest createQuestionRequest) {
+
     Question question = convertCreateQuestionRequestToQuestion(createQuestionRequest);
+
+    if (!questionService.isQuestionRequestValid(question)) {
+      throw new QuestionException("You can not save invalid question as finished!");
+    }
```
stanislavgrujic, line 52: Actually, we want to allow the user to save an invalid question, but do not want to take it into account once we generate a formula, or a test based on that formula.

web/src/main/java/eu/execom/testgenerator/web/util/QuestionException.java
```diff
@@ -0,0 +1,9 @@
+package eu.execom.testgenerator.web.util;
+
+public class QuestionException extends RuntimeException {
```
stanislavgrujic, line 3: This is very generic and not very informative. It is better to have more specific exceptions which are informative (in this case, if we would want to forbid saving incomplete questions it would be `IncompleteQuestionException`)

Issue #490, opened 2020-06-15
Tests fail during build


VojislavKovacevic:
**Tests fail during build**

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running eu.execom.testgenerator.core.service.category.CategoryServiceImplTest
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.27 sec - in eu.execom.testgenerator.core.service.category.CategoryServiceImplTest
Running eu.execom.testgenerator.core.service.question.QuestionServiceTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.031 sec - in eu.execom.testgenerator.core.service.question.QuestionServiceTest
Running eu.execom.testgenerator.core.service.test.ScheduledExpiredTestsHandlerTest
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/C:/Users/vkovacevic/.m2/repository/ch/qos/logback/logback-classic/1.1.11/logback-classic-1.1.11.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/C:/Users/vkovacevic/.m2/repository/org/slf4j/slf4j-log4j12/1.7.25/slf4j-log4j12-1.7.25.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.109 sec - in eu.execom.testgenerator.core.service.test.ScheduledExpiredTestsHandlerTest
Running eu.execom.testgenerator.core.service.test.TestServiceTest
Tests run: 10, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.752 sec <<< FAILURE! - in eu.execom.testgenerator.core.service.test.TestServiceTest
testApplyReview_whenNotAllReviewed_expectNoAchievedPoints(eu.execom.testgenerator.core.service.test.TestServiceTest)  Time elapsed: 0.007 sec  <<< ERROR!
java.lang.NullPointerException: null
[...]

PR #504, opened 2020-06-19
Tests fail during build #490

adamovivan:


core/src/main/java/eu/execom/testgenerator/core/service/email/EmailServiceImpl.java
```diff
@@ -12,6 +12,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.mail.SimpleMailMessage;
 import org.springframework.mail.javamail.JavaMailSender;
+import org.springframework.scheduling.annotation.Async;
```
stanislavgrujic, line 15: Please remove unused import.

Issue #496, opened 2020-06-16
Minimum number of questions for valid formula


adamovivan:
The formula is considered valid if it has a minimum of 10 questions. This check should be done when questions, recipes, and formula create, update and delete operations are performed.

PR #505, opened 2020-06-19
 Minimum number of questions for valid formula #496 

adamovivan:


core/src/main/java/eu/execom/testgenerator/core/service/formula/FormulaServiceImpl.java
```diff
@@ -131,10 +131,17 @@ public void updateFormulaValidity(Question question) {
...
   private boolean isValid(Formula formula) {
-    return formula.getRecipes().stream().allMatch(this::hasEnoughQuestions);
+    if (formula.getRecipes().isEmpty()) {
+      return false;
+    }
+
+    boolean categoryHasEnoughQuestions = formula.getRecipes().stream().allMatch(this::categoryHasEnoughQuestions);
+    int totalQuestions = formula.getRecipes().stream().mapToInt(Recipe::getNumberOfQuestions).sum();
+
+    return categoryHasEnoughQuestions && totalQuestions >= 10;
```
stanislavgrujic, line 141: Please make a number of needed questions a matter of configuration, instead of hardcoding it.

core/src/main/java/eu/execom/testgenerator/core/util/Constants.java
```diff
@@ -2,12 +2,15 @@
...
+  private Constants() {
+  }
 
   public static final String PATH_API_USERS = "/api/users";
 
   public static final String PATH_ACCEPT_INVITE = "/accept-invite";
 
   public static final String PATH_ACCEPT_TEST_INVITE = "/accept-test-invite";
 
+  public static final int MINIMUM_NUMBER_OF_QUESTIONS = 10;
```
stanislavgrujic, line 14: This is not a configuration, but a constant. Configuration would be if you had extracted it as a property and make it possible to configure the minimum number of questions without having to change the source code and rebuild and redeploy the application.

adamovivan: Should I put it in the application.properties instead?

stanislavgrujic: Yes, please do that and make sure to update the example file.

PR #519, opened 2020-06-24
PR from last week which was merged with 474 branch, but needs to be merged with develop.

draganaHrcek:
PR from last week which was merged with 474 branch, but needs to be merged with develop.


web/src/main/java/eu/execom/testgenerator/web/controller/QuestionController.java
```diff
@@ -41,8 +42,15 @@ public QuestionController(QuestionService questionService) {
   }
 
   @PostMapping
-  public ResponseEntity<CreateQuestionResponse> createNewQuestion(@RequestBody @Valid CreateQuestionRequest createQuestionRequest) {
+  public ResponseEntity<CreateQuestionResponse> createNewQuestion(
+      @RequestBody @Valid CreateQuestionRequest createQuestionRequest) {
+
     Question question = convertCreateQuestionRequestToQuestion(createQuestionRequest);
+
+    if (!questionService.isQuestionRequestValid(question)) {
```
stanislavgrujic, line 50: Instead of having to check each time before you save a question if it is valid, you should make method `isQuestionRequestValid` private and call it from the `save` method itself. That way you will make sure that it is always invoked, and you do not rely on somebody knowing *the protocol* for saving question (check if it is valid, then save).

web/src/main/java/eu/execom/testgenerator/web/util/IncompleteQuestionException.java
```diff
@@ -0,0 +1,9 @@
+package eu.execom.testgenerator.web.util;
+
+public class IncompleteQuestionException extends RuntimeException {
+
+  public IncompleteQuestionException(String message) {
```
stanislavgrujic, line 5: Instead of passing the same message 3 times (whenever you create this exception), you could create a default constructor and set the message inside it.

stanislavgrujic:
I have accidentally approved, instead of requesting changes. Please fix them before merging.

core/src/test/java/eu/execom/testgenerator/core/service/question/QuestionServiceTest.java
```diff
@@ -93,15 +97,49 @@ public void shouldCreateNew() {
...
+  @Test(expected = IncompleteQuestionException.class)
+  public void updateShouldThrowException() {
+
+    allQuestions.get(0).setFinished(true);
+    allQuestions.get(0).setType(QuestionType.MULTIPLE_CHOICE);
+
+    Answer answer = new Answer();
+    answer.setCorrect(false);
+
+    ArrayList<Answer> answers = new ArrayList<Answer>();
```
stanislavgrujic, line 121: Please use an interface (`List<Answer>`) to declare a variable.
Please use a diamond operator to implicitly declare the type, on the right side of the assignment operator.
`List<Answer> answers = new ArrayList<>();`

core/src/test/java/eu/execom/testgenerator/core/service/question/QuestionServiceTest.java
```diff
@@ -93,15 +97,50 @@ public void shouldCreateNew() {
...
+  @Test(expected = IncompleteQuestionException.class)
+  public void saveShouldThrowException() {
+
+    allQuestions.get(0).setFinished(true);
+    allQuestions.get(0).setType(QuestionType.MULTIPLE_CHOICE);
+
+    Answer answer = new Answer();
+    answer.setCorrect(false);
+
+    List<Answer> answers = new ArrayList<>();;
```
stanislavgrujic, line 139: Please remove surplus semi-colon.

Issue #419, opened 2020-04-22
Elements overlapping on the review screen 


VojislavKovacevic:
Elements overlapping on the review screen. For question type "coding challenge", the dropdown menu is not fully visible.

![elementsIn reviewe2020-04-22 11_40_25-Execom test](https://user-images.githubusercontent.com/63228397/79978605-79201180-84a0-11ea-99ff-9388c267ad0d.png)

VojislavKovacevic:
Issue still present. Develop branch 15.6.2020.

![test overlaping2020-06-15 11_16_59-Execom test](https://user-images.githubusercontent.com/63228397/84640906-9d292d80-aefa-11ea-9a87-dbe41b0a1683.png)

adamovivan:
Please check it again, it should work. There is another problem: 
![Screenshot from 2020-06-23 15-34-03](https://user-images.githubusercontent.com/29520040/85410277-2d085080-b567-11ea-8179-77ce04a43c7b.png)

PR #526, opened 2020-06-24
Line numbers overlapping over code #419 

adamovivan:


web/src/main/frontend/src/components/ReviewCodeChallengeQuestion.js
```diff
@@ -10,6 +10,17 @@ import 'codemirror/mode/jsx/jsx';
 import 'codemirror/mode/php/php';
 
 class ReviewCodeChallengeQuestion extends Component {
+  componentDidMount() {
+    setTimeout(
```
aberonja, line 14: setTimeout? 🤔

adamovivan: There is some problem with the CodeMirror component. I tried I lot of things and nothing worked. Some people on the internet solved the problem in that way.

aberonja: Please, leave a comment then why you did it like this.

adamovivan: I leave some time for everything to initialize completely, so then trigger change state. After the state is changed CodeMirror bug disappears.

PR #528, opened 2020-06-24
[525-517] REVIEWER_ASSIGNED test status added, reviewer assignment fo…

draganaHrcek:
REVIEWER_ASSIGNED test status added, invite reviewer forbidden for STARTED, REVIEW_IN_PROGRESS, REVIEW_FINISHED tests. Also reviewer can see only REVIEWER_ASSIGNED, REVIEW_IN_PROGRESS, REVIEW_FINISHED tests.


core/src/test/java/eu/execom/testgenerator/core/service/test/TestServiceTest.java
```diff
@@ -395,25 +398,128 @@ public void shouldAssignReviewer() {
...
+  public void shouldNotAssignReviewer_TestWithStatusSTARTED() {
+    // Given
+    Test test = forSubmission();
+    test.setStatus(TestStatus.STARTED);
+    Long testId = test.getId();
+
+    when(testRepository.getOne(testId)).thenReturn(test);
+
+    //When
+    boolean result= testService.assignReviewerToTest(testId, any());
```
stanislavgrujic, line 468: Please format the source code properly.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -115,7 +116,11 @@ public TestController(TestService testService, JwtTokenManager jwtTokenManager,
 
   @PutMapping(value = "/actions/assign-reviewer")
   public ResponseEntity<Void> assignReviewer(@RequestBody AssignReviewerRequest assignReviewer) {
-    testService.assignReviewerToTest(assignReviewer.getTestId(), assignReviewer.getReviewerId());
+
+    if (!testService.assignReviewerToTest(assignReviewer.getTestId(), assignReviewer.getReviewerId())) {
```
stanislavgrujic, line 120: This is part of the business logic and it should be thrown from the method itself, not in the place where you invoke the method. The way you use it, each person invoking `assignReviewerToTest` should know that it is needed to throw an exception if you cannot assign reviewer. Please move logic inside the mehtod.

web/src/main/java/eu/execom/testgenerator/web/util/ReviewerCanNotBeAssignedException.java
```diff
@@ -0,0 +1,9 @@
+package eu.execom.testgenerator.web.util;
+
+public class ReviewerCanNotBeAssignedException extends RuntimeException {
+
+  public ReviewerCanNotBeAssignedException(String message) {
```
stanislavgrujic, line 5: Same as my comment in other PR - if you are not passing different messages to the exception, you should set the message internally.

core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -287,9 +287,17 @@ public void applyReview(Test test, TestReview review) {
 
   @Override
   @Transactional
-  public void assignReviewerToTest(Long testId, Long reviewerId) {
-
+  public boolean assignReviewerToTest(Long testId, Long reviewerId) {
```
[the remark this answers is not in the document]
draganaHrcek: I didn't change method for tests, a changed it because I was throwing an exception in controller. I made it like this because other controllers in project are implemented like that.(https://prnt.sc/t68l7s) In my experiences I also put all business logic in services not in endpoint's. Won't throw exceptions in controllers anymore :)

core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -288,8 +283,19 @@ public void applyReview(Test test, TestReview review) {
...
+
+    switch (test.getStatus()) {
+      case STARTED:
+      case REVIEW_FINISHED:
+      case REVIEW_IN_PROGRESS:
+        throw new ReviewerCanNotBeAssignedException();
+      case REVIEWER_ASSIGNED:
+        break;
+      default:
+        test.setStatus(TestStatus.REVIEWER_ASSIGNED);
```
stanislavgrujic, line 296: It is a good practice to add `break` at the end of` default` case as well. Please update the code.

Issue #522, opened 2020-06-24
Delete user


VojislavKovacevic:
- Administrators should have the option to remove users.  
- After removal, users will not be visible from the application but should be preserved in DB. 
- Tests assigned to removed users, in statuses "REVIEWER_ASSIGNED" and "REVIEW_IN_PROGRESS" should be changed to status "FINISHED". 
- Allocated points and comments if any, should be removed. This action should be in the "Users" tab, accessible to admin only.

VojislavKovacevic:
The functionality will not be developed. It is not needed, user which do not work in Execom anymore will be prevented to log in anyway.

PR #529, opened 2020-06-25
 Delete user #522 

adamovivan:


web/src/main/frontend/src/scss/components/_user.scss
```diff
@@ -100,9 +100,13 @@
           .user__edit_icon {
             background-image: url('img/edit_white.png');
           }
+          .user__trash_icon {
```
aberonja, line 103: \n

web/src/main/frontend/src/scss/components/_user.scss
```diff
@@ -120,6 +124,9 @@
         .user__edit_icon {
           background-image: url('img/edit.png');
         }
+        .user__trash_icon {
```
aberonja, line 127: \n

web/src/main/frontend/src/components/UserTableBody.js
```diff
@@ -10,6 +10,10 @@ class UsersTableBody extends React.Component {
     this.props.user.activateChangeRolePopup();
   };
 
+  rowDelete = row => () => {
```
aberonja, line 13: Does this need to be `= row => () => {}`? Could it be `= row => {}`?

aberonja: Also, you can pass id since you don't need whole row object.

core/src/main/java/eu/execom/testgenerator/core/service/user/UserServiceImpl.java
```diff
@@ -59,7 +62,26 @@ public User update(User user) {
   @Override
   @Transactional
   public void remove(Long id) {
-    userRepository.delete(id);
+    User user = userRepository.findById(id);
+
+    if (user.isReviewer()) {
+      for (Test test: user.getReviewerTests()) {
```
stanislavgrujic, line 68: This would look nicer if you have used Stream API:
```
user.getReviewerTests().stream()
       .filter(this::isPendingReview)
       .forEach(this::restoreToFinished);

private boolean isPendingReview(Test test) {
  return test.getStatus() == TestStatus.REVIEWER_ASSIGNED || test.getStatus() == TestStatus.REVIEW_IN_PROGRESS;
}

private void restoreToFinished(Test test) {
  test.setReviewer(null);
  test.setStatus(TestStatus.FINISHED);

  for (TestQuestion testQuestion: test.getTestQuestions()) {
    testQuestion.setAchievedPoints(null);
    testQuestion.setComment(null);
  }

  test.setComment(null);
}

```

As a side note, if you have many places where you you check the values of an enumeration, it is a good practice to internalize those checks. E.g., in this case, you could have created a method in TestStatus:
```
public boolean isPendingReview() {
  return this == REVIEWER_ASSIGNED || this == REVIEW_IN_PROGRESS;
}
```

This way, you will not miss a place if your conditions change.

PR #534, opened 2020-06-26
Change test status in REVIEW_IN_PROGRESS

draganaHrcek:


web/src/main/frontend/src/scss/components/_reviewTest.scss
```diff
@@ -205,3 +205,7 @@ textarea {
 .review-index {
   z-index: 1;
 }
+
+.cancel_button {
+  float: left !important;
```
aberonja, line 210: Please replace this with flex if we're using flex on this project.

draganaHrcek: Is it ok now? :)

Issue #531, opened 2020-06-25
Comment field has fixed size


adamovivan:
On some smaller screen resolutions points field won't be visible

PR #536, opened 2020-06-26
Comment field has fixed size #531 

adamovivan:


web/src/main/frontend/src/scss/components/_reviewTest.scss
```diff
@@ -187,21 +187,27 @@ textarea {
...
 
 .review-question-comment-div {
-  z-index: 5;
-  margin-top: -29px;
+  width: 90%;
+  margin-left: 2%;
+}
+
+.points-container {
+  width: 8%;
```
aberonja, line 205: All those width properties look sketchy. Please take few more moments to see if there's a better way before merging.

PR #539, opened 2020-06-26
 change reviewer

draganaHrcek:
On change of reviewer, mail needs to be sent to previous and new reviewers, informing them about change.


core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -286,11 +286,16 @@ public void assignReviewerToTest(Long testId, Long reviewerId) {
     Test test = findOne(testId);
 
     switch (test.getStatus()) {
+      case CREATED:
       case STARTED:
       case REVIEW_FINISHED:
       case REVIEW_IN_PROGRESS:
         throw new ReviewerCanNotBeAssignedException();
       case REVIEWER_ASSIGNED:
+        if(test.getReviewer()!=null && test.getReviewer().getId()==reviewerId){
```
stanislavgrujic, line 295: This is not formatted properly, spaces are missing.
Also, you are comparing objects (id) by using `==`, instead of using `equals`.

core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -286,11 +286,16 @@ public void assignReviewerToTest(Long testId, Long reviewerId) {
...
+         return;
+        }
+        emailService.reviewerHasBeenChangedNotification(test.getReviewer().getEmail(), test);
```
stanislavgrujic, line 298: Please pay attention to warnings from IDE. Here it says that you might get a `NullPointerException` on `test.getReviewer()`.

core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -286,11 +286,16 @@ public void assignReviewerToTest(Long testId, Long reviewerId) {
...
         break;
```
stanislavgrujic, line 299: In general, I think it is a bad place for adding sending notification - switch case, before the change, was being used to check if test is in proper status and change it accordingly, and now you have coupled it with logic to send email notification. This will make it very difficult to extract this piece of code to a method. 
Also, logic for sending notifications is now scattered in the method, instead of having it nicely grouped in a single place. My suggestion is to think how could you update the method code so that sending notifications is done in a single place (see line 314, where `emailService.inviteReviewer()` is invoked) .
Another problem is that it can happen that you send notification about the user being changed, but the logic for test update fails. To prevent this, you should send a notification only after you are sure that the update was saved successfully (again, see line 314).

core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -284,13 +284,21 @@ public void applyReview(Test test, TestReview review) {
...
 
     switch (test.getStatus()) {
+      case CREATED:
       case STARTED:
       case REVIEW_FINISHED:
       case REVIEW_IN_PROGRESS:
         throw new ReviewerCanNotBeAssignedException();
       case REVIEWER_ASSIGNED:
+        if(test.getReviewer() != null){
+          if (test.getReviewer().getId().equals(reviewerId)){
```
stanislavgrujic, line 297: It was more readable with && for both conditions, than with nested, please revert it.

core/src/main/java/eu/execom/testgenerator/core/service/test/TestServiceImpl.java
```diff
@@ -307,6 +315,10 @@ public void assignReviewerToTest(Long testId, Long reviewerId) {
     test.setReviewer(user);
     testRepository.save(test);
     emailService.inviteReviewer(user.getEmail());
+
+    if (oldReviewerEmail!=null){
+      emailService.reviewerHasBeenChangedNotification(oldReviewerEmail, test);
+    }
   }
```
stanislavgrujic, line 322: Formatting is still not good - there should be a space between a keyword (`if` or any other) and open parenthesis. Before/after any parenthesis/brackets there should be a space character. Around comparison signs, assignment, calculation signs, there should be a space character. Could you please make sure that your IDE is configured to do so automatically? Feel free to contact me to do it together, if you are struggling with it.

Issue #351, opened 2018-12-03
Preview submitted test


vrackovic:
- There should be a option for previewing both solved and reviewed tests.
- New endpoint should be created
- New view should be created and triggered with click on row in `Dashboard` component

PR #551, opened 2020-06-30
 Preview submitted test #351 

adamovivan:


web/src/main/frontend/src/base/stores/DashboardStore.js
```diff
@@ -19,7 +18,9 @@ class DashboardStore {
         row.awardedPoints !== null ? row.awardedPoints : '?';
       return {
         ...row,
-        points: `${awardedPoints}/${row.maxPoints} ${Math.ceil(
+        points: `${awardedPoints}/${
+          row.maxPoints
+        }\u00A0\u00A0\u00A0\u00A0\u00A0${Math.ceil(
```
aberonja, line 23: Why does this looks like this?

web/src/main/frontend/src/components/ReviewTestQuestionComponent.js
```diff
@@ -46,15 +53,22 @@ class ReviewTestQuestionComponent extends React.Component {
...
-              questionComment={this.props.answer.comment}
-            />
+            <span
+              hidden={
+                !this.props.completedTest.isReviewed &&
+                !this.props.completedTest.reviewActive
+              }
+            >
+              <Points
+                id={this.props.question.id}
```
aberonja, line 63: Why do you need an ID?

web/src/main/frontend/src/base/stores/DashboardStore.js
adamovivan: It's a non-breaking space used inside of template literal because I need more spacing between points and percents.

web/src/main/frontend/src/components/ReviewTestQuestionComponent.js
adamovivan: It is needed for determining which points belong to which question when we sending entered points to the backend.

PR #552, opened 2020-06-30
Switching to new dto  for Save action, old one contains validations for Finish action.

draganaHrcek:


web/src/main/java/eu/execom/testgenerator/web/dto/SaveReviewChangesDto.java
```diff
@@ -0,0 +1,19 @@
+package eu.execom.testgenerator.web.dto;
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.Map;
+
+@Data
+@NoArgsConstructor
+public class SaveReviewChangesDto {
```
stanislavgrujic, line 10: Could you please be consistent in naming - the other class is named `SubmitReviewRequest`. Could you name this one `SaveReviewRequest`? It is very difficult to understand the code if we are not using the similar naming patterns (e.g., how do you determine what is a request and what is a dto?)

Issue #547, opened 2020-06-30
Empty error  message


VojislavKovacevic:
If the reviewer comment is longer then 256 characters, review cant be finished (correct).  
An empty error message will popup (incorrect).

PR #554, opened 2020-06-30
Comment empty error message #547 

adamovivan:


web/src/main/java/eu/execom/testgenerator/web/validator/MaxCommentLength.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Constraint(validatedBy = MaxCommentLengthValidator.class)
+@Target({ElementType.METHOD, ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface MaxCommentLength {
+
+  String message() default "Maximum comment length has been exceeded";
+  Class<?>[] groups() default {};
+  Class<? extends Payload>[] payload() default {};
+  int max() default 0;
```
stanislavgrujic, line 18: With the default value being used, it will always be invalid. It is a good design choice to use such value that with the default value being used, it should pass the validation. My suggestion is that if it is set to default value, 0, it is always valid, no matter the length.

stanislavgrujic: This is still not handled. I see you have made a change to address it, but later on reverted it. Do you plan to add another fix?

adamovivan: Yes, sorry, my mistake

PR #562, opened 2020-07-06
User should not be administrator and reviewer at the same time.

draganaHrcek:


web/src/main/frontend/src/components/ChangeRole.js
```diff
@@ -6,16 +6,47 @@ import SelectRole from '../components/SelectRole';
...
   }
 
-  changeHandler = roles => {
+  changeHandler = (roles, lastSelected) => {
     if (!roles) {
       return;
     }
 
-    this.props.changeRoleForm.set({roleNames: roles.map(role => role.value)});
+    var newRoles = [];
```
aberonja, line 22: Var ❌

web/src/main/frontend/src/components/ChangeRole.js
```diff
@@ -6,16 +6,47 @@ import SelectRole from '../components/SelectRole';
...
+      roles.map(role => role.label).includes('Administrator')
+    ) {
+      for (let role of roles) {
+        if (role.label !== 'Administrator') {
+          newRoles.push(role);
+        }
+      }
+    } else {
+      newRoles = roles;
+    }
```
aberonja, line 45: \n

PR #572, opened 2020-07-10
Generate PDF document for tests in status CREATED, FINISHED, REVIEW_FINISHED

draganaHrcek:


web/src/main/frontend/src/base/stores/DashboardStore.js
```diff
@@ -100,6 +100,17 @@ class DashboardStore {
     });
   }
 
+  @action.bound
+  saveTestPdf(id) {
+    testApi.saveTestPdf(id).then(response => {
+      let url = window.URL.createObjectURL(response.xhr.response);
+      let a = document.createElement('a');
```
aberonja, line 107: Try const, it should work.

web/src/main/frontend/src/components/DashboardTableBody.js
```diff
@@ -24,6 +24,10 @@ class DashboardTableBody extends React.Component {
     this.props.dashboard.enableReviewerPopup(row.reviewer);
   };
 
+  saveTestPdf = row => {
+    return this.props.dashboard.saveTestPdf(row.id);
+  };
```
aberonja, line 29: `saveTestPdf = row => this.props.dashboard.saveTestPdf(row.id)`

core/src/main/java/eu/execom/testgenerator/core/service/jasperReport/SimpleReportFiller.java
```diff
@@ -0,0 +1,93 @@
...
+    try {
+      InputStream reportStream = getClass().getResourceAsStream("/template/jasperTemplates/".concat(reportFileName));
+      jasperReport = JasperCompileManager.compileReport(reportStream);
+      File file = new File(System.getProperty("user.dir") + File.separator + "jasperReports");
+      file.mkdir();
+      JRSaver.saveObject(jasperReport,
+          System.getProperty("user.dir") + File.separator + "jasperReports" + File.separator + reportFileName.replace(
+              ".jrxml", ".jasper"));
+    } catch (JRException ex) {
+      Logger.getLogger(SimpleReportFiller.class.getName()).log(Level.SEVERE, null, ex);
```
stanislavgrujic, line 54: Please use class level `@Slf4j` annotation and log as is done throughout the code base. Also, if you just log the error, how will the end user know that there has been an error while the report was being generated?

core/src/main/java/eu/execom/testgenerator/core/service/jasperReport/SimpleReportFiller.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  public void fillReport() {
+    try {
+      jasperPrint = JasperFillManager.fillReport(jasperReport, parameters, dataSource.getConnection());
+    } catch (JRException | SQLException ex) {
+      Logger.getLogger(SimpleReportFiller.class.getName()).log(Level.SEVERE, null, ex);
```
stanislavgrujic, line 62: Please use class level `@Slf4j` annotation and log as is done throughout the code base. Also, if you just log the error, how will the end user know that there has been an error while the report was being generated?

core/src/main/java/eu/execom/testgenerator/core/service/jasperReport/SimpleReportFiller.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.File;
+import java.io.InputStream;
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Component
+public class SimpleReportFiller {
```
stanislavgrujic, line 22: This will not work in a multiuser environment where multiple users can generate reports simultaneously. You should **never** keep the state in shared (Spring) components!
My suggestion is to remove field reportFileName and pass the report name as parameter to `compileReport` method.

core/src/main/java/eu/execom/testgenerator/core/service/jasperReport/SimpleReportFiller.java
```diff
@@ -0,0 +1,93 @@
...
+    return reportFileName;
+  }
+
+  public void setReportFileName(String reportFileName) {
+    this.reportFileName = reportFileName;
+  }
+
+  public JasperPrint getJasperPrint() {
+    return jasperPrint;
+  }
```
stanislavgrujic, line 92: You should not have public getters and setters in Spring beans, since it shows that you are using them to store state. Since these components are shared (by default created as a singleton instance) you will never know what the actual state is in multithreaded (multiuser) environment, nor will have the guarantee of proper execution.

web/src/main/java/eu/execom/testgenerator/web/controller/TestController.java
```diff
@@ -96,6 +100,11 @@ public TestController(TestService testService, JwtTokenManager jwtTokenManager,
     return ResponseEntity.ok(new TestDto(test));
   }
 
+  @GetMapping(value = "pdfReport/{testId}", produces = MediaType.APPLICATION_PDF_VALUE)
```
stanislavgrujic, line 103: This is not very restful. I suggest to update the endpoint so that is matches - `GET {serverUrl}/api/tests/{testId}/report`

core/src/main/java/eu/execom/testgenerator/core/service/jasperReport/SimpleReportFiller.java
```diff
@@ -50,44 +43,18 @@ public void compileReport() {
...
   }
 
-  public void fillReport() {
+  public JasperPrint fillReport(Map<String, Object> parameters) {
     try {
       jasperPrint = JasperFillManager.fillReport(jasperReport, parameters, dataSource.getConnection());
-    } catch (JRException | SQLException ex) {
-      Logger.getLogger(SimpleReportFiller.class.getName()).log(Level.SEVERE, null, ex);
+    } catch (JRException | SQLException e) {
+      log.error("Could not fill report with data", e);
```
adamovivan, line 55: How will the end-user know that there has been an error while the report was being generated?

adamovivan:
Please resolve the conflicts.

PR #585, opened 2020-07-17
Problem with deleted questions during testing solved. (and removed unused  method)

draganaHrcek:


core/src/main/java/eu/execom/testgenerator/core/service/question/QuestionService.java
```diff
@@ -15,7 +14,5 @@
 
   void remove(Long id);
 
-  List<Question> findAllByCategory(Category category);
-
```
adamovivan, line 19: Check if it is used in web project

